PR #65, opened 2016-05-08
[readme] refresh_session -> enable-refresh-tokens

rvion:


rvion:
@gambol99 there is an other mention here: https://github.com/rvion/keycloak-proxy/blob/patch-1/README.md#--the-default-config

gambol99:
@rvion needs to be rebased with master ...

Issue #68, opened 2016-05-08
docker hub images are not up-to date


rvion:
https://hub.docker.com/r/gambol99/keycloak-proxy/tags/

rvion:
for some reasons, docker-hub still says “Last pushed: 3 months ago"

gambol99:
the builds go to https://quay.io/repository/gambol99/keycloak-proxy ...

rvion:
arf x) 

I think a mention in the readme would be great :)

regarding the current image on hub.docker.com, how about either 
  1. deleting the image on docker hub
  2. manually pushing new images from time to time (like for major releases)
  3 pushing to both quay and docker hub

(Thanks for the quick answers, as always !)

gambol99:
There is a badge at the top, but i'll look at pushing to both ... for some reason though even when I do a successful docker login, docker.io keeps throwing a "authentication required" error

PR #71, opened 2016-05-08
trais build

gambol99:
- adding a travis build


gambol99:
in references to issue: [#68](https://github.com/gambol99/keycloak-proxy/issues/68)

Issue #64, opened 2016-05-07
stuck in an infinite loop


rvion:
with keycloak-proxy compiled from master, I got  stuck in a infinite redirection cycle
(tested with keycloak 1.9.1.Final)

```
 keycloak-prooxy-mater \
    --discovery-url=https://auth.rvion.fr/auth/realms/master/.well-known/openid-configuration \
    --listen=:3000 \
    --client-id=unprotectedservice \
    --upstream-url=http://unprotectedservice:8080 \
    --redirection-url=http://keycloakproxy:3000 \
    --client-secret=secret \
    --resource="uri=/" \
    --verbose=true
```

when I access http://keycloakproxy:3000, I'm redirected to keycloak. Then, I login on keycloak, and I'm redirected back to http://keycloakproxy:3000/oauth/callback?xxx. But then, it enter in a redirection loop, and chrome shows me an error message

I see in the logs:

```
ERRO[0016] failed to get session, redirecting for authorization  error=authentication session not found
INFO[0016] incoming authorization request from client address: 37.161.218.27:41982  access_type= client_ip=37.161.218.27:41982
ERRO[0019] failed to get session, redirecting for authorization  error=authentication session not found
INFO[0019] incoming authorization request from client address: 37.161.218.27:41982  access_type= client_ip=37.161.218.27:41982
ERRO[0057] failed to get session, redirecting for authorization  error=authentication session not found
INFO[0057] incoming authorization request from client address: 37.161.218.27:41982  access_type= client_ip=37.161.218.27:41982
INFO[0058] issuing a new access token for user, email: vion.remi@gmail.com  duration=58.995194544s email=vion.remi@gmail.com expires=07 May 16 20:55 +0000 idle=0
ERRO[0058] failed to get session, redirecting for authorization  error=authentication session not found
INFO[0058] incoming authorization request from client address: 37.161.218.27:41982  access_type= client_ip=37.161.218.27:41982
INFO[0058] issuing a new access token for user, email: vion.remi@gmail.com  duration=59.588883831s email=vion.remi@gmail.com expires=07 May 16 20:55 +0000 idle=0
ERRO[0058] failed to get session, redirecting for authorization  error=authentication session not found
INFO[0058] incoming authorization request from client address: 37.161.218.27:41982  access_type= client_ip=37.161.218.27:41982
[...]

PR #72, opened 2016-05-08
config invalid for secure-cookie

gambol99:
- the configuration should be invalid if the redirection-url i.e. the site is non-ssl but the secure-cookie option is set true


gambol99:
In reference to issue #64

PR #104, opened 2016-06-20
docker user

gambol99:
- shifting the version to v1.2.1
- adding a user in the docker file to ensure we dont run as root

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,16 @@ FROM alpine:3.3
 MAINTAINER Rohith <gambol99@gmail.com>
 
 RUN apk update && \
-    apk add ca-certificates
+    apk add ca-certificates && \
+    adduser -S proxy && \
+    addgroup -S proxy
 
 ADD templates/ opt/templates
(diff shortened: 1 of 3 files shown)


gambol99:
fixes issue https://github.com/gambol99/keycloak-proxy/issues/88

gambol99:
@timgent?

timgent:
Thanks!

nabadger:
@gambol99 

Hi - it looks like this isn't on current master anymore - did it get removed for any reason?

I can see it was removed in #107 

We have a policy on our k8s cluster that doesn't allow containers to run as root (so I'd have to build my own container to use this). 

I'm not sure there's a good reason to run any container as root these days - there always tends to be available workarounds right?

Thanks

jangaraj:
@nabadger I guess, you can use [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/):
```
  securityContext:
    runAsUser: 1000
    fsGroup: 2000
```

Issue #112, opened 2016-07-08
Forward signing proxy... for client certs


timgent:
Just an idea - forward signing proxy can do keycloak auth at the min, could configure it to also do auth with client certificates

gambol99:
Just to be clear ... you want an extra option to use a client certificate in the TLS client config and 'perhaps' reload the certificate when the client cert comes close to expiration?

timgent:
Yeah. What do you think? Not sure if it is OTT or not.

gambol99:
@timgent add in PR #114

timgent:
Nice :)

PR #114, opened 2016-07-13
client certificate

gambol99:
- adding the command line options -tls-client-ceritificate to load a client cert for upstream forwarding and reverse proxy


gambol99:
addresses issue #112

Issue #117, opened 2016-07-28
Cross compile and add to releases


chrisns:
Hey,
We've got a user who can't use docker, and is on windows
Could you therefore also have some cross compiles to different archs and have travis put them in the release artifacts?
I need something like:
```
GOOS=windows GOARCH=amd64 ./bin/godep go build -a -o bin/keycloak-proxy-amd64.exe
GOOS=windows GOARCH=386 ./bin/godep go build -a -o bin/keycloak-proxy-i386.exe
```

PR #118, opened 2016-07-29
cross compile to windows in release

chrisns:
addresses #117
see https://github.com/chrisns/keycloak-proxy/releases/tag/no for example release

@gambol99, you'll need to add your github oauth token in the `.travis.yml`

I also tidied up so that it does the docker build+push, it was running some other make job?
And also made it only try and do deploy bits when it's running Go 1.6, otherwise it would deploy twice


gambol99:
LGTM

PR #123, opened 2016-08-17
Res code failed auth

timgent:
Now uses a response code of 401 for failed auths instead of 500

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -24,9 +24,11 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/coreos/go-oidc/oauth2"
 	"github.com/gin-gonic/gin"
 )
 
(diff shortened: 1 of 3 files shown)


handlers.go
```diff
@@ -19,11 +19,13 @@ import (
 	"bytes"
 	"encoding/base64"
 	"fmt"
+	"github.com/coreos/go-oidc/oauth2"
```
gambol99: So a go idiom is local packages at top, local package in the middle and external at the bottom

oauth_test.go
```diff
@@ -76,6 +76,9 @@ Ka0WPQGKjQJhZRtqDAT3sfnrEEUa34+MkXQeKFCu6Yi0dRFic4iqOYU=
 -----END RSA PRIVATE KEY-----
 `
 
+const validUsername = "test"
```
gambol99: const (
   validUsername = "test"
   validPassword = "test"
)

gambol99:
other than above ...lgtm

Issue #136, opened 2016-09-28
cookie-domain bug


boyand:
Currently the cookie-domain flag is set as a StringSlice. Therefore if you pass ```--cookie-domain foo.com``` on startup, dropping a cookie fails with:

```net/http: invalid Cookie.Domain "[foo.com]"; dropping domain attribute```

Using a config file works correctly.

gambol99:
hi @boyand ... cheers for spotting this one, I've raised https://github.com/gambol99/keycloak-proxy/pull/137 which should fix the issue

PR #137, opened 2016-09-28
cookie domain command line opts bug

gambol99:
- fixing the --cookie-domain option
- adding extra test for verify the cookie domain is working correctly


gambol99:
related to bug #136

Issue #146, opened 2016-11-07
Support Identity provider with self-signed certificate


johanneslanger:
@gambol99  First of all thanks for your great work on this project :)

In our test environment we have a keycloak installation which has a self signed certificate. The proxy fails to start when retrieving the discovery URL:

`
_time="2016-11-07T10:10:51Z" level=warning msg="failed to get provider configuration from discovery url: https://<keycloak-url>/auth/realms/<realm>, Get https://<keycloak-url>/auth/realms/<realm>/.well-known/openid-configuration: x509: certificate signed by unknown authority"
`

I would suggest to add a config option `--skip-openid-provider-tls-verify` to allow a self-signed certs for test environments!

If you want I can try and contribute

PR #147, opened 2016-11-07
Adding config option to skip tls verify for OpenID provider

johanneslanger:
This is in response to issue #146

Did this fix today so we are unblocked. Would love to get this in. Let me know what you think!

The config param defaults to false as this could be a security issue if set to true in production.
Did not find a quick way to write test for this though :(

Let me know what you think!

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -115,6 +115,8 @@ type Config struct {
 	RedirectionURL string `json:"redirection-url" yaml:"redirection-url"`
 	// RevocationEndpoint is the token revocation endpoint to revoke refresh tokens
 	RevocationEndpoint string `json:"revocation-url" yaml:"revocation-url"`
+	// SkipOpenIDProviderTLSVerify skips the tls verification for openid provider communication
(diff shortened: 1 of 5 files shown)


doc.go
```diff
@@ -115,6 +115,8 @@ type Config struct {
 	RedirectionURL string `json:"redirection-url" yaml:"redirection-url"`
 	// RevocationEndpoint is the token revocation endpoint to revoke refresh tokens
 	RevocationEndpoint string `json:"revocation-url" yaml:"revocation-url"`
+	// skips the tls verification for openid provider communication
```
gambol99: Needs the variable name added to the comment ...as it's exported (go-vet will complain) .. i.e. 
// SkipOpenIDProviderTLSVerify skip ...

johanneslanger:
Ok, updated the comment to include the variable name!

gambol99:
thank you for the fix @johanneslanger :-)

PR #173, opened 2017-01-09
Fix typo in docs that actually affects the usage :)

easternbloc:
Due to auto magic from your doc generation you will actually have to set the env var of `PROXY_CLIENT_SECERT` which is sadly a typo.

This fixes that ❤️


gambol99:
LGTM

PR #183, opened 2017-01-26
Redirection URL

gambol99:
- removed the need to specify a redirection-url and using either the Host header or X-Forwarded-Host (if present)


gambol99:
fixes issue https://github.com/gambol99/keycloak-proxy/issues/182

Issue #187, opened 2017-02-01
Ajax redirects and being logged out of the proxied app


cameronbraid:
I'm using keycloak proxy in front of an installation of grafana  and one of prometheus.

I can login just fine, but since most of the interaction between the browser and the server is via ajax, once the access token has expired, future ajax calls receive a redirect to the authorize page.  Both of these two apps (grafana and prometheus) don't have any script to catch this redirect, so you end up being logged out and have to reload the page (or open the homepage in a tab) to automatically log back in.

I understand that keycloak and keycloak-proxy support refresh tokens, but from my experimentation they still require the browser to interact with keycloak to refresh the tokens.  

I was hoping that keycloak proxy could refresh the tokens automatically when proxying ajax requests.  Is this possible ?

gambol99:
So the issue is related to the access token cookie expiring too quickly. The problem is caused [here](https://github.com/gambol99/keycloak-proxy/blob/master/middleware.go#L118); if the access token is removed, even with a kc-state (i.e. refresh token) it never gets to the section where the token will be refreshed. At the moment, I've placed a configuration option to control the access token cookie expiration, though I'm not sure of the best practice here. I could set the expiration = refresh token expiration, but not all Idp's hand have a readable refresh token format (google for example), perhaps attempt to get refresh token expiration and if unable default to the configuration option is a compromise ... any thought are most welcome? :-)

cameronbraid:
Ok, great - that looks like it will do the job.

gambol99:
fixed in #189

gambol99:
closing as fixed

PR #188, opened 2017-02-01
Access Token Duration

gambol99:
- fixed the access token expiring too quickly; the cookie expiration needed to be extended.
  Not sure when this was broken, but I changed the duration a while back. Needs a test added.


gambol99:
fixed #187

Issue #200, opened 2017-03-17
Make multiple slashes equivalent to a single one in urls


bogbert:
Hi,
I'm running keycloak-proxy 2.0.3 behind an nginx reverse proxy.
nginx is configured to redirect the location /app/soap to keycloak-proxy
The keycloak-proxy yml configuration file declares a single resource:
```
resources:
- uri: /app/soap
  roles:
  - app:access_webservices
```
I noticed that if a user tries to access the url //app/soap or /app//soap, then nginx redirects it to keycloak-proxy, but keycloak-proxy considers that it doesn't match /app/soap, and authentication is completely skipped !
Just like nginx, I would like keycloak-proxy to consider multiple slashes in urls to be equivalent to a single one.

For now, as a workaround, I declared a resource that denies access to all non-matching urls:
```
- uri: /
  roles:
  - deny_role
```
where deny_role is a fake role that is not assigned to anyone.
Would there be a more elegant way to deny access by default to non-matching urls ?

gambol99:
That's not good at all!! ... There doesn't appear to be an easy way to get rid of this either as the HTTP router doesn't provide any options to massage the urls in gin or pure http router. Perhaps introducing middleware to filter urls is the easier.

```go
// filterMiddleware is custom filtering for incoming requests
func (r *oauthProxy) filterMiddleware() gin.HandlerFunc {
	return func(cx *gin.Context) {
		var p rune
		var b bytes.Buffer
		for _, c := range cx.Request.URL.Path {
			if c == '/' && p == '/' {
				continue
			}
			p = c
			b.WriteRune(c)
		}
		cx.Request.URL.Path = b.String()
	}
}
````

https://play.golang.org/p/qWpT4HZ_FD

gambol99:
fixed in #201

PR #201, opened 2017-03-17
bugfix

gambol99:
FIXES:
 * Fixes a bug in authentication, which permitted double slashed url entry [#PR200](https://github.com/gambol99/keycloak-proxy/pull/200)

FEATURES:
 * Grabbing the revocation-url from the idp config if user override is not specified [#PR193](https://github.com/gambol99/keycloak-proxy/pull/193)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 
 #### **2.0.4**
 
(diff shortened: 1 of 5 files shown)


gambol99:
related to #200

PR #202, opened 2017-03-18
Normalize URL

gambol99:
- normal the url before we apply the protection middleware against it


gambol99:
looking at normalizing all urls before applying the middleware ... related to #200

PR #207, opened 2017-04-20
Run as a non root user

easternbloc:
This isn't tested but I think it's all that's required to get the container running as a non root user.

Let me know if I've missed anything.


Dockerfile
```diff
@@ -6,8 +6,12 @@ RUN apk update && \
 
 ADD templates/ opt/templates
 ADD bin/keycloak-proxy /opt/keycloak-proxy
-RUN chmod +x /opt/keycloak-proxy
+RUN useradd keycloak && \
```
gambol99: It's alpine linux, you would probably need something like 
```shell
addgroup -S keycloak
adduser -G keycloak -S keycloak
```

Dockerfile
```diff
@@ -6,8 +6,12 @@ RUN apk update && \
...
+    chmod +x /opt/keycloak-proxy && \
+    chown keycloak:keycloak /opt/keycloak-proxy
```
gambol99: this line shouldn't be required, as it should already been executable

easternbloc:
@gambol99 I notice you changed all the things since I did this. Let me know if you're happy with that approach!

easternbloc:
@gambol99 pingy pong!

PR #234, opened 2017-06-04
[*] Fix #233 store encrypted refresh token

### LINKED ISSUE #233: refresh token stored without encryption
If a store is enabled, the refresh token will be stored  in plain text other than encrypted token after refreshing new token, this will cause error like ` in plain text other than encrypted token`

https://github.com/gambol99/keycloak-proxy/blob/master/middleware.go#L219

antmanler:


gambol99:
cheers for the fix @antmanler ... lgtm

PR #249, opened 2017-07-11
Add support for LetsEncrypt certificates, closes #231

### LINKED ISSUE #231: letsencrypt support
Add LetsEncrypt support for dynamic certs

nl5887:
* added config param use-letsencrypt to enable letsencrypt
* added code to request letsencrypt certificate

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -238,7 +238,7 @@ type Config struct {
 	UseLetsEncrypt bool `json:"use-letsencrypt" yaml:"use-letsencrypt" usage:"use letsencrypt for certificates"`
 
 	// LetsEncryptCacheDir is the path to store letsencrypt certificates
-	LetsEncryptCacheDir string `json:"letsencrypt-cache-dir" yaml:"lets-encrypt-cache-dir" usage:"path where cached letsencrypt certificates are stored"`
+	LetsEncryptCacheDir string `json:"letsencrypt-cache-dir" yaml:"letsencrypt-cache-dir" usage:"path where cached letsencrypt certificates are stored"`
 
 	// SignInPage is the relative url for the sign in page
 	SignInPage string `json:"sign-in-page" yaml:"sign-in-page" usage:"path to custom template displayed for signin"`
--- a/doc.go
+++ b/doc.go
@@ -234,6 +234,12 @@ type Config struct {
 	// EnableProxyProtocol controls the proxy protocol
 	EnableProxyProtocol bool `json:"enabled-proxy-protocol" yaml:"enabled-proxy-protocol" usage:"enable proxy protocol"`
 
+	// UseLetsEncrypt controls if we should use letsencrypt to retrieve certificates
+	UseLetsEncrypt bool `json:"use-letsencrypt" yaml:"use-letsencrypt" usage:"use letsencrypt for certificates"`
+
+	// LetsEncryptCacheDir is the path to store letsencrypt certificates
+	LetsEncryptCacheDir string `json:"letsencrypt-cache-dir" yaml:"lets-encrypt-cache-dir" usage:"path where cached letsencrypt certificates are stored"`
+
 	// SignInPage is the relative url for the sign in page
 	SignInPage string `json:"sign-in-page" yaml:"sign-in-page" usage:"path to custom template displayed for signin"`
 	// ForbiddenPage is a access forbidden page
--- a/config.go
+++ b/config.go
@@ -40,6 +40,8 @@ func newDefaultConfig() *Config {
 		SecureCookie:                true,
 		SkipUpstreamTLSVerify:       true,
 		SkipOpenIDProviderTLSVerify: false,
+		UseLetsEncrypt:              false,
+		LetsEncryptCacheDir:         "./cache/",
 	}
 }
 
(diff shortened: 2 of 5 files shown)


nl5887:
In second thought, probably better to use hostnames argument instead of redirectionUrl to validate valid hostnames for certificates. Do you agree?

doc.go
```diff
@@ -234,6 +234,12 @@ type Config struct {
 	// EnableProxyProtocol controls the proxy protocol
 	EnableProxyProtocol bool `json:"enabled-proxy-protocol" yaml:"enabled-proxy-protocol" usage:"enable proxy protocol"`
 
+	// UseLetsEncrypt controls if we should use letsencrypt to retrieve certificates
+	UseLetsEncrypt bool `json:"use-letsencrypt" yaml:"use-letsencrypt" usage:"use letsencrypt for certificates"`
+
+	// LetsEncryptCacheDir is the path to store letsencrypt certificates
+	LetsEncryptCacheDir string `json:"letsencrypt-cache-dir" yaml:"lets-encrypt-cache-dir" usage:"path where cached letsencrypt certificates are stored"`
```
nl5887: I have also use-letsencrypt, should we go for all letsencrypt or lets-encrypt? use-lets-encrypt seems a bit strange.

gambol99:
hi @nl5887 ... the problem with using the hostnames or redirectionURL is both of these are optional variables. If no redirectionURL is set it will us the Host header and hostnames is used exclusively in the security [middleware](https://github.com/gambol99/keycloak-proxy/blob/master/middleware.go#L356-L357). I guess the easiest compose is to use the redirectionURL if there or default to all? ...

Also, i'm assuming this would only work with a single instance no?, given the [cache](https://github.com/golang/crypto/blob/master/acme/autocert/cache.go) is directory based? ..

nl5887:
We could add an extra letsencrypt-hostnames or make hostnames mandatory. Other solution could be to just default to true if no hostnames are set. Reason to have this options is due to the "Failed Validation" limits.

nl5887:
It should work on multiple instances, as long as the rate limit will not be reached. So multiple instances using same hostname will request their own certificate and each individual hostname will have its own certificate in the caching dir.

gambol99:
hi @nl5887 ... sorry for delay ... 
> We could add an extra letsencrypt-hostnames or make hostnames mandatory. Other solution could be to just default to true if no hostnames are set. Reason to have this options is due to the "Failed Validation" limits.

So i'd be reluctant to make --hostnames enforceable, one neat feature of not requiring it means you can use the proxy layer above a numerous of sites to provide auth. I guess sticking to how it's implemented now makes senses? ... 
```go 
if len(hostnames) > 0 
   it must be one of those in []hostnames
else if a redirectionURL != "", 
   must be Host
end

otherwise neither and we assume return true ... 
```

gambol99:
LGTM ... much appreciated @nl5887!!

jangaraj:
Thank you guys for LE feature,

Is it possible to add some LE example into README pls?

My config:
```
...
enable-https-redirection: true
enable-security-filter: true
use-letsencrypt: true
letsencrypt-cache-dir: /lecache
redirection-url: https://domain.tld:443/
hostnames:
  - domain.tld
...
```

LE cache:
```
$ ls -lah /lecache
total 4
drwxr-xr-x    2 root     root          30 Jul 25 21:36 .
drwxr-xr-x   19 root     root         277 Jul 25 21:34 ..
-rw-------    1 root     root         227 Jul 25 21:34 acme_account.key
```

Testing:
```
$ openssl s_client -connect domain.tld:443 -servername domain.tld
CONNECTED(00000003)
140584098252440:error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error:s23_clnt.c:769:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 7 bytes and written 339 bytes
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : 0000
    Session-ID:
    Session-ID-ctx:
    Master-Key:
    Key-Arg   : None
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1501018949
    Timeout   : 300 (sec)
    Verify return code: 0 (ok)
---
```

Logs:
```
1.501018453182229e+09   info    starting the service    {"prog": "keycloak-proxy", "author": "Rohith", "version": "v2.1.0-rc2 (git+sha: 0eadf33, built: 25-07-2017)"}
1.501018453182799e+09   info    attempting to retrieve configuration discovery url      {"url": "https://accounts.google.com"}
1.5010184532526886e+09  info    successfully retrieved openid configuration from the discovery
1.5010184532605608e+09  info    enabled reverse proxy mode, upstream url        {"url": "http://localhost:8081/"}
1.5010184532609131e+09  info    enabling the security filter middleware
1.5010184532610123e+09  info    enabled the service metrics middleware, available on    {"path": "/oauth/metrics"}
1.5010184532613084e+09  info    protecting resource     {"resource": "uri: /*, methods: GET, required: authentication only"}
1.5010184532616882e+09  info    keycloak proxy service starting {"interface": "0.0.0.0:443"}
```

Any idea what's missing in my case/config?

gambol99:
cc @nl5887 ... also, i'm gonna copy and paste this into an issue for tracking

PR #264, opened 2017-08-18
use --no-cache

Vad1mo:
--update isn't available in alpine 3.6 its better to use --no-cache this always downloads the latest index


gambol99:
LGTM ... thank you kindly @Vad1mo

PR #266, opened 2017-08-23
Correct typo and added ENV Variable

Vad1mo:
There is an copy and past error in the documentation. PROXY_ENABLE_SECURITY_FILTER is currently mapped to enable-refresh-tokens which is not correct this means that enable-refresh-tokens does not have an env variable so it was added.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -143,9 +143,9 @@ type Config struct {
 	// EnableForwarding enables the forwarding proxy
 	EnableForwarding bool `json:"enable-forwarding" yaml:"enable-forwarding" usage:"enables the forwarding proxy mode, signing outbound request"`
 	// EnableSecurityFilter enabled the security handler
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -143,9 +143,9 @@ type Config struct {
 	// EnableForwarding enables the forwarding proxy
 	EnableForwarding bool `json:"enable-forwarding" yaml:"enable-forwarding" usage:"enables the forwarding proxy mode, signing outbound request"`
 	// EnableSecurityFilter enabled the security handler
-	EnableSecurityFilter bool `json:"enable-security-filter" yaml:"enable-security-filter" usage:"enables the security filter handler"`
+	EnableSecurityFilter bool `json:"enable-security-filter" yaml:"enable-security-filter" usage:"enables the security filter handler"` env:"ENABLE_SECURITY_FILTER"`
```
gambol99: need to remove the ` after handler

Vad1mo:
fixed issue!

PR #268, opened 2017-08-28
HTTP Server and Upstream Timeouts

gambol99:
Related to issue [263](https://github.com/gambol99/keycloak-proxy/issues/263). Permitting the users to set varioues timeouts on the http.Server and upstream proxy


gambol99:
sorry for the delay @jangaraj ... can you review please?

jangaraj:
LGTM 👍 . Note: default non zero values for server read/write timeouts can be breaking change, because default Golang zero values are used at the moment = no timeouts. Probably it'll be good to mention it in breaking changes section. 

Thanks!

gambol99:
good point ... i'll put it into the changelog!

PR #272, opened 2017-09-06
Proxy for IDP + README update

jangaraj:
1.) Fix for https://github.com/gambol99/keycloak-proxy/issues/271
New config option for proxy, which will be used for communication with IDP (that's common use case in the enterprise environment)

2.) A few typos in the README


config.go
```diff
@@ -40,6 +40,7 @@ func newDefaultConfig() *Config {
 		ServerReadTimeout:           5 * time.Second,
 		ServerWriteTimeout:          10 * time.Second,
 		SkipOpenIDProviderTLSVerify: false,
+		OpenIDProviderProxy:         "",
```
gambol99: I think we can drop that one .. it's gonna default to "" anyhow; though if you wanna keep it for informational purposes can we order them alphabetically (i'm slight ocd on that :-))

server.go
```diff
@@ -605,6 +605,18 @@ func (r *oauthProxy) newOpenIDClient() (*oidc.Client, oidc.ProviderConfig, *http
 	// step: create a idp http client
 	hc := &http.Client{
 		Transport: &http.Transport{
+			Proxy: func(_ *http.Request) (*url.URL, error) {
+				if r.config.OpenIDProviderProxy != "" {
```
gambol99: Out of interest what's the reason behind using is via an option rather than relying on ```Proxy: http.ProxyFromEnvironment```

config.go
jangaraj: Done, removed.

server.go
jangaraj: This implementation is explicit proxy configuration, and it provides more flexibility for configuration: config file, command line option, (env variable).

gambol99:
lgtm

PR #275, opened 2017-09-14
Improve docs in README

ernsheong:


ernsheong:
Oops looks like edits have to come via https://github.com/gambol99/keycloak-proxy/blob/master/doc.go

PR #279, opened 2017-09-20
Use divided cookies for large access/refresh token cookie - fix #278

### LINKED ISSUE #278: Problem with kc-access/kc-state cookie, which has more than 4093 bytes
As a user I have a lot of assigned groups, so my access token can have more than 4093 bytes. Also, token encryption adds some additional bytes. The problem is that (encrypted) access/refresh token is stored in the cookie (`kc-access/kc-state`) and the browser may have a problem to write more than 4093 bytes per single cookie - it will result in the infinite loop of issuing access token between keycloak-proxy and IDP because cookie value is not available.

Solution:
If the length of cookie data for the `kc-access` cookie is larger than 4093 bytes, then set more cookies in the browser, (`kc_access, kc-access-1, kc-access-2, ...`) so the single cookie will store max 4093 bytes. Also when keycloak-proxy reads `kc-access` cookie, it will read all cookies with prefix `kc-access_` and it will combine them into one value internally.  

Reference for the constant `4093` - http://browsercookielimits.squawky.net/.

Any better ideas how to solve this browser cookie size limit?

WIP: https://github.com/gambol99/keycloak-proxy/compare/master...jangaraj:divided-access-cookie
image

jangaraj:


gambol99:
apologizes for the delay @jangaraj ... i'll take a look at this this evening

session.go
```diff
@@ -100,8 +102,25 @@ func getTokenInBearer(req *http.Request) (string, error) {
...
 func getTokenInCookie(req *http.Request, name string) (string, error) {
+	var token bytes.Buffer
+
 	if cookie := findCookie(name, req.Cookies()); cookie != nil {
-		return cookie.Value, nil
+		token.WriteString(cookie.Value)
+	}
+
+	// add also divided cookies
+	for i := 1; i < 600; i++ {
```
gambol99: we could just use the len() of the cookies rather than arbitrary number

session.go
```diff
@@ -100,8 +102,25 @@ func getTokenInBearer(req *http.Request) (string, error) {
...
+		cookie := findCookie(name+"-"+strconv.Itoa(i), req.Cookies())
+		if cookie == nil {
+			break
+		} else {
+			token.WriteString(cookie.Value)
+		}
+	}
+
+	if token.Len() == 0 {
```
gambol99: we can jump the if else and us 
```go
if token.Len() == 0 {
   return "", ErrSessionNotFound
}

return token.String(), nil
```

session.go
jangaraj: Yes, `len(req.Cookies())` is better option.

session.go
jangaraj: Done.

gambol99:
hi @jangaraj ... i'm guessing your hitting a newer version of the go-resty ...  You need to use glide to install the deps

```shell
$ go get github.com/Masterminds/glide
$ glide install 
$ make test
```

jangaraj:
Yes, I use glide and go-resty version defined in `glide.lock`:
```
$ glide install
...
[INFO]  --> Found desired version locally github.com/go-resty/resty 39c3db9c7bb4f9718ac143a83a924441521caf73!
...
[INFO]  --> Setting version for github.com/go-resty/resty to 39c3db9c7bb4f9718ac143a83a924441521caf73.
...
$ make test
--> Running the tests
# _/root/git/keycloak-proxy
./middleware_test.go:155:11: request.SetProxy undefined (type *resty.Request has no field or method SetProxy)
FAIL    _/root/git/keycloak-proxy [build failed]
make: *** [test] Error 2
```

gambol99:
@jangaraj ... very odd! .. Given this is a required feature though, i'll merge and add some tests ..

PR #284, opened 2017-10-30
Run the proxy as a non-root user

RobinKnipe:
Add new user `kcproxy` to avoid running as the `root` user.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,11 @@ RUN apk add --no-cache ca-certificates
 ADD templates/ /opt/templates
 ADD bin/keycloak-proxy /opt/keycloak-proxy
 
+RUN useradd -mUrd /kcproxy kcproxy \
+    && chown -R kcproxy:kcproxy /opt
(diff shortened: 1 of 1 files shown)


jangaraj:
Hi Robin,

My personal notes: 

I don't like hardcoding of any usernames into base Docker image. See `grafana/grafana-docker` issues - some users have problems with mounting of config files, data files, different user IDs in the and outside of the container. For many users, it's too complicated and confused.

A better option is to have environment variable, where a user can define which user will be created/used in the container (not in the image). The best option is to use [user namespaces](https://docs.docker.com/engine/security/userns-remap/), so `root` in the container is not a `root` on your Docker host.

RobinKnipe:
Ah, it actually turns out this has already been done in `gitlab`, and it's that version (with a `keycloack` user) that gets used to build the image, which is then pushed to quay.io/ukhomeofficedigital/go-keycloak-proxy:v2.1.0-rc2

Issue #285, opened 2017-11-15
prevent HTTP 413


pschulten:
@gambol99 Thanks for this project.
Is it possible to not send the Cookie Header to upstream?
In my current use-case I have no control on upstream and it has a very conservative header size policy :( 
With encryption and refresh tokens the Cookie header is too large for upstream and it sends an 413

jangaraj:
How do your cookie sizes look like in the browser? There was my issue and my PR, which increase the effective size of data in the cookies - https://github.com/gambol99/keycloak-proxy/issues/278 - Do you use latest master branch version?

pschulten:
Thanks for this feature @jangaraj . I saw that and compiled from latest master branch but that didn't help in my case. The Cookie payload was around 3.8k. I guess my upstream has a hard header limit of 4k for all headers combined. I fixed it by stripping my JWT payload to the absolute minimum but It would be nice if the keycloak-proxy provides a switch to drop the openidc cookies on reverse proxying.

jangaraj:
I agree. I have had a similar issue, but I was able to increase header size limit in my infrastructure. My suggestion:
```
cookie-blacklist:
   - kc_access*
   - kc_state*
```

gambol99:
hi @pschulten ... i've started a PR to add the functionality #287 .. just need to fix up the unit test but i should have this available by the weekend

gambol99:
fixed in #287

jangaraj:
Note about potential cookie suffix: https://github.com/gambol99/keycloak-proxy/commit/81c7892b906ae65edbf9a62d65e375084bc44f8d

gambol99:
good spot @jangaraj !!!! ... i'll amend the check and update the unit test ... thank you kindly :-)

pschulten:
@gambol99 thank you very much. It works like a charm!
@jangaraj thanks for your support

songokudbz:
Where did you add the cookie-blacklist?

songokudbz:
Ty! Works. Nice solution.

PR #286, opened 2017-11-17
Upstream Authorization Cookie

gambol99:
- adding an option to stop the proxy from including the authorization cookies in the upstream request

so disabling all authorization tokens would mean specifying `--enable-authorization-cookies=false --enable-authorization-header=false --enable-token-header=false` .. 

TODO: add shortcut to disable all threw


gambol99:
fixes #285

Issue #280, opened 2017-09-26
App context/vhosts domains in the upstream config


jangaraj:
As a developer I want to use keycloak-proxy as a standard reverse proxy. That means I want to use app context in the upstream setting, for example: https://domain.com/app-context/ or I want to use it with the domain, where virtual hosts are enabled (`Host` header needs to be part of the upstream request).

Will it be possible?

pjeby:
As of 2.1.0, you can now set the Host: header in your `headers:`.

jangaraj:
Sure, it's partial improvement. But some legacy apps need also context (`/something`).

pjeby:
Probably that means prefixing `req.URL.Path` with the `r.endpoint.Path` in that same bit of code.  I don't actually have a use-case for this at the moment, so I'm not currently motivated to slog through the escaping and other issues.  I did find an example of reverse proxy code [here](https://blog.charmes.net/post/reverse-proxy-go/), and translating what it does comes out to:

```go
req.URL.Path = singleJoiningSlash(r.endpoint.Path, req.URL.Path)
```

But I am uncertain what happens to any URL escaping in the paths, since this is manipulating `Path` and not `RawPath`.  Close reading of the `net/url/url.go` source suggests that the above is *probably* correct, but it *will* alter the form of some escaped URLs.

(The code sample I linked also included some query string manipulation (to add query params from the reverse proxy target); that code would *probably* work here as well; at least it manipulates encoded paths.  But I'm not clear on the use case for always adding query params, vs. the one for a path prefix.)

Anyway, you can try adding that code to the source and see if it works for you.

pjeby:
Hm.  By the way, it looks like `X-Forwarded-Host` was supposed to be the original URL host, but the way the forwarding code is written, it's been sending the target host instead.  The `req.Header` manipulations should be before the URL manipulation.

kurt---:
indeed the X-Forwarded-Host-value is set wrong. This breaks clients (e.G. my symfony-app) which rely on this value. Hope this could be fixed in the next release. Thanks for this great software!

abstractj:
@jangaraj thanks for taking your time reporting this. The repository keycloak-proxy was moved to Keycloak organization and renamed to keycloak-gatekeeper. If you still think this issue is valid, please read https://www.keycloak.org/community.html and report your bug/feature request to https://issues.jboss.org/browse/KEYCLOAK.

Issue #259, opened 2017-08-02
Unable to set the request Host header


slaash:
Hi,

In our setup the keycloak proxy's upstream-url is set to the internal URL of a web application that is not accessible from the internet (e.g. http://webapp.internal:6543/).
We use an Amazon Load Balancer that forwards requests to the keycloak proxy. It is accessed publicly like https://protected-webapp.example.com.

Our issue is thet the webapp relies on the Host header from the http request for creating URLs, so now we are getting the internal hostname (webapp.internal:6543) in the browser, which of course does not work from the internet.

I see that the Host header is set by the keycloak proxy to the value of upstream-url.
I tried to set the Host header manually like this:
headers:
  Host: protected-webapp.example.com
But still it gets to the webapp with the upstream-url value.

Is there any way to set the Host header to a custom value?

Thanks!

songokudbz:
Hi I've also have kinda the same problem.

I have 2 containers in a Kubernetes POD. One is mapped to PORT 8080 (the actual application) and another container with the keycloak-proxy thing. The thing is if I try to access the internal container with trailing slash ('/'), i.e. <PUBLIC_URL>/<CONTEXTPATH>**/**, then after the login on keycloak side I am being sent back to the <PUBLIC_URL>/<CONTEXTPATH>/ and everything works. BUT, if I go to the URL without the trailing slash, i.e. <PUBLIC_URL>/<CONTEXTPATH>, then I will be redirected back from Keycloak after a successful login to http://localhost:8080/<CONTEXTPATH> which is not good.

How can I solve this?

jangaraj:
@songokudbz try to set config `redirection-url="<PUBLIC_URL>"`

songokudbz:
@jangaraj it did not do the trick. My upstream url looks like this:
upstream-url: "http://localhost:{{ service.ports[0].targetPort }}"
do I need to change it as well to the PUBLIC URL?

jangaraj:
@songokudbz: No, you don't need to update `upstream-url`. I think that redirect to the localhost is generated by your app, not by keycloak-proxy -> debug/fix your app code/config.

pjeby:
As of 2.1.0, you can now set the Host: header in your `headers:`.

gambol99:
Closing the issue as it was fixed in v2.1.0

PR #290, opened 2017-11-28
Allow setting upstream Host header

pjeby:
See #259 and #280.  This change allows a custom upstream `Host:` header to be set, for use cases where the internal and external URIs need to be the same due to legacy constraints, but DNS for the host resolves to the proxy rather than the upstream.


gambol99:
apologizes @pjeby for the delay .. i only just saw this one ..

gambol99:
apologizes @pjeby for the delay .. i only just saw this one ..

PR #297, opened 2017-12-21
Go Packages

gambol99:


abstractj:
@gambol99 do you think we can close this? I noticed that these changes are a bit dated when compared to master.

gambol99:
@abstractj ... agreed!! ... I should have merged ages ago really .. fixing it up now would be more hassle than redoing it from scratch :-)

PR #298, opened 2017-12-21
Authentication Cookies

gambol99:
- need to use a prefix rather than just the name due to the cookie being split


abstractj:
@gambol99 same here. Please file a Jira (https://issues.jboss.org/browse/KEYCLOAK), in this way we can prioritize stuff for the next release.

Issue #300, opened 2018-01-05
Matching dex groups via claim match


ginkel:
This is more of a question than an issue.

I'm trying to get keycloak-proxy b73e8bd (current master) running against dex (with an LDAP backend) as an oidc provider. dex does not know about roles, but supports a `groups` claim, which may end up with a token like this:

```
{
  "iss": "https://sso.fqdn",
  "sub": "CgVzbW9rZRIEbGRhcA",
  "aud": "my-app",
  "exp": 1515269245,
  "iat": 1515182845,
  "at_hash": "...",
  "email": "me@privacy.net",
  "email_verified": true,
  "groups": [
    "app-group"
  ],
  "name": "John Doe"
}
```

Is there a way in keycloak-proxy to express a claim match that only permits access if the group `app-group` is present in the `groups` claim?

Thanks!

gambol99:
Hi @ginkel ... I've raised a quick PR #301 which implements access controls on the `groups` claim as it's a useful feature and I know Keycloak now includes this claim as well. One thing, unlike the `roles` where the user requires all roles specified, groups are applied with an OR operation. This in my head made sense, but let me know your thoughts.

ginkel:
Excellent, works like a charm! Thanks! :-)

My use-cases are currently so simple that I typically only have a single group per resource, but I'm wondering why the behavior should be different from that of other claims that all need to be satisfied for an access to be permitted.

PR #301, opened 2018-01-06
Groups Claim

gambol99:
The current implementation does not take into account the standard `groups` claim when matching on the resource. This PR add the ability to add an additional field `--groups=list,of,groups` to the resource.

Note unlike the roles which are applied with an AND operation a user simply has to exist in one of the groups specified.


gambol99:
Adds the ability to caters for the `groups` claim within a token and implements #300

PR #303, opened 2018-01-07
Websocket Fix

gambol99:
The http2FancyWriter does not implement the http.Hijacker method causing the websocket upgrade to throw an exception. This PR fixes the issues by forcing the selection of httpFancyWriter{}


aalubin:
tested and works like a charm

gambol99:
awesome! ... thanks for testing @aalubin :-)

PR #311, opened 2018-02-07
Websocket Headers

gambol99:
- fixing up the headers for websockets


gambol99:
fixes the issue raised in https://github.com/gambol99/keycloak-proxy/issues/310

PR #315, opened 2018-02-23
OpenID Provider Configuration Timeout

gambol99:
- adding the ability to control the timeout on the openid configuration pull


gambol99:
provides the feature request in https://github.com/gambol99/keycloak-proxy/issues/307

PR #328, opened 2018-03-08
Added --preserve-host option to retain host header in upstream request.

jgroffen:
Useful for upstream webapps that rely on the host header to generate weblinks that match the host originally requested.

Similar to the Apache mod_proxy ProxyPreserveHost option: https://httpd.apache.org/docs/trunk/mod/mod_proxy.html#proxypreservehost


jgroffen:
Rebased from master to resolve merge conflicts (and keep this PR to a single commit).

forwarding.go
```diff
@@ -50,7 +50,7 @@ func (r *oauthProxy) proxyMiddleware(next http.Handler) http.Handler {
 		if v := req.Header.Get("Host"); v != "" {
 			req.Host = v
 			req.Header.Del("Host")
-		} else {
+		} else if !r.config.PreserveHost {
```
gambol99: hi @jgroffen ... so flag is used to control the Host header from the incoming request be preserved for upstream, which the if statement just above is doing. .. perhaps i'm missing something?

jgroffen: Hello @gambol99, I found that without this patch all code paths would change the host header of the upstream request - either to the configured upstream host name or an overridden header in the headers config. Maybe there was something screwy with my config though, I'll mark it as WIP and check.

jgroffen: Hello @gambol99 - I'm finding that this if statement on line 50 is never equating to true - I don't know why because the incoming request does have a host header.

I verified the following configuration:

    listen : kcproxy.internal.com
    upstream-url : webserver.internal.com

... when keycloak-proxy is behind a public-facing proxy with a request chain that looks like:

* **public.com** --proxy to--> **kcproxy.internal.com**  --proxy to--> **webserver.internal.com**

... produces the following results:

1. request to public.com host header: **public.com**
1. request to kcpublic.internal.com host header: **public.com**
1. request to webserver.internal.com host header: **webserver.internal.com**

Note that **PreserveHost** is enabled for the **public.com** proxy - an Apache HTTPD proxy in this case.

I verified this result using the 2.2.1 released binary. I also checked a request with a valid token vs a request that has no auth yet and performs OIDC redicrect to keycloak (where SSO happens).

If the preserve-host parameter is set in my changed codebase, the following happens instead:

1. request to public.com host header: **public.com**
1. request to kcpublic.internal.com host header: **public.com**
1. request to webserver.internal.com host header: **public.com**

We made a change where we don't need to preserve the host header anymore (why I took so long to get back to you, sorry :( ). Preserving host headers is not a preferred approach, but some applications rely on it.

PR #335, opened 2018-03-27
Optional login

### LINKED ISSUE #270: The case of login being optional
On a classic website you have a front page.
On this front page you have something like a "Welcome Firstname Lastname" or "Login or Register" if not logged in.

When defining resources that need to be guarded, to my knowledge, there is no possibility to say "login is optional".
```yaml
discovery-url: https://domain/auth/realms/testrealm
client-id: someid
client-secret: somesecret
listen: 127.0.0.1:8080
encryption_key: 1234567890123456
upstream-url: http://127.0.0.1:3000
secure-cookie: false
resources:
- uri: /doc*
  methods:
  - GET
  - POST
  - DELETE
```

This is a test program
```go
package main

import (
	"fmt"
	"net/http"
	"strings"

	"github.com/SermoDigital/jose/jws"
)

func main() {
	h := &Handler{}
	http.ListenAndServe(":3000", h)
}

type Handler struct{}

func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	switch r.Method {
	case http.MethodGet:
		fmt.Fprintln(w, "GET")
		h.printXHeaders(w, r)
		j, err := jws.ParseFromHeader(r, jws.Unknown)
		if err != nil {
			fmt.Fprintln(w, err.Error())
			return
		}
		fmt.Fprintln(w, j.Payload())
		i, _ := j.Payload().(map[string]interface{})
		fmt.Fprintln(w, i["name"])
	case http.MethodPost:
		fmt.Fprintln(w, "POST")
		h.printXHeaders(w, r)
	case http.MethodDelete:
		fmt.Fprintln(w, "DELETE")
		h.printXHeaders(w, r)
	default:
		fmt.Fprintln(w, "NOT ALLOWED")

	}
}

func (h *Handler) printXHeaders(w http.ResponseWriter, r *http.Request) {
	for k, v := range r.Header {
		if strings.HasPrefix(k, "X-") {
			fmt.Fprintln(w, k, v)
		}
	}
}
```

with the above program if you *are* logged in and hit `/` uri the result is

```
GET
X-Forwarded-For [127.0.0.1]
X-Forwarded-Host [127.0.0.1:3000]
X-Forwarded-Proto []
no token present in request
```

So I'm suggesting to have a configurable means to pass the token regardless of an endpoint or endpoint prefix being protected or not.

If you already can do that then sorry for bothering.

pjeby:
This patch from @justjanne addresses the issue of sites with pages that need to behave differently depending on whether a user is actually logged in.  Specifically, it:

Transmit auth headers even for whitelisted resources

- Let whitelisted resources also be handled by authentication and
  headers middleware, but not by admission middleware

- Refactor authentication middleware into a function that returns the
  type of response that is expected (authenticateRequest), and one
  that acts on it (authenticationMiddleware)

- Handle the responseTypes RedirectAuthentication and AccessForbidden
  the same as Serve if the resource is whitelisted

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -95,7 +95,7 @@ func (r *oauthProxy) authenticateRequest(w http.ResponseWriter, req *http.Reques
 	user, err := r.getIdentity(req)
 	if err != nil {
 		r.log.Error("no session found in request, redirecting for authorization", zap.Error(err))
-		return responseTypeRedirectAuthentication, nil
+		return responseTypeRedirectAuthentication, req.Context()
 	}
 	// create the request scope
 	scope := req.Context().Value(contextScopeName).(*RequestScope)
@@ -214,7 +214,9 @@ func (r *oauthProxy) authenticationMiddleware(resource *Resource) func(http.Hand
 			if resource.WhiteListed {
 				switch responseType {
 				case responseTypeServe:
+					fallthrough
 				case responseTypeRedirectAuthentication:
+					fallthrough
 				case responseTypeAccessForbidden:
 					next.ServeHTTP(w, req.WithContext(ctx))
 					return
--- a/middleware.go
+++ b/middleware.go
@@ -83,10 +83,10 @@ func (r *oauthProxy) loggingMiddleware(next http.Handler) http.Handler {
 }
 
 const (
-	responseTypeServe = 0
+	responseTypeServe                  = 0
 	responseTypeRedirectAuthentication = 1
-	responseTypeAccessForbidden = 2
-	responseTypeError = 3
+	responseTypeAccessForbidden        = 2
+	responseTypeError                  = 3
 )
 
 func (r *oauthProxy) authenticateRequest(w http.ResponseWriter, req *http.Request) (int, context.Context) {
--- a/middleware.go
+++ b/middleware.go
@@ -82,131 +82,162 @@ func (r *oauthProxy) loggingMiddleware(next http.Handler) http.Handler {
 	})
 }
 
+const (
+	responseTypeServe = 0
+	responseTypeRedirectAuthentication = 1
+	responseTypeAccessForbidden = 2
+	responseTypeError = 3
+)
+
+func (r *oauthProxy) authenticateRequest(w http.ResponseWriter, req *http.Request) (int, context.Context) {
+	clientIP := req.RemoteAddr
+	// grab the user identity from the request
+	user, err := r.getIdentity(req)
+	if err != nil {
+		r.log.Error("no session found in request, redirecting for authorization", zap.Error(err))
+		return responseTypeRedirectAuthentication, nil
+	}
+	// create the request scope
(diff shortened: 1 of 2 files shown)


gambol99:
hi @pjeby .. sorry, i'm not ignoring this :-) ... but the repo might be moving over to keycloak org thus I'm not merged anything too big for now ..

dalu:
However it doesn't take the various methods into account, that is if that's the same version that's mentioned in #270 therefor it's "not good enough".

justjanne:
@dalu my own use case allows anonymous create and read, but not anything else (think like imgur with authenticated/anonymous uploads).

The goal of my patch for #270 was to allow that. Letting the app require authentication for whatever it wants, but if authentication headers are available, they are sent anyway.

dalu:
The question is, is the different from the version in #270?
If not then it doesn't work like that.

pjeby:
I've changed the issue title to no longer reference #270 so that this PR can stand on its own.  It is, after all, quite functional for both my and @justjanne's use case of "app handles its own security for whitelisted URLs".

@dalu: Looking at your notes on #270, it's not entirely clear to me what problem you're saying you're having.  You posted a configuration, a log, and a crash message, none of which I am able to interpret other than the configuration.  You didn't say what the log *should* have looked like, and you also mentioned something about using a different version of resty.  Given that, and the fact that @justjanne's patch was "works for me", I went ahead and submitted this PR.

Can you please describe clearly what you want this patch to do that it does not do?   "Doesn't work properly" is not very helpful or informative.  All I have understood so far is that it has something to do with different rules for different methods on the same URL, but I don't know what you expect it to do or what it's actually doing instead, so there is no way for me to make it "work properly" without knowing either of those things.

bogdanpetrea:
Hey, what's the status of this PR?

pjeby:
"Should do" is relative.  It works fine for the use case of globally-optional login, and has done so in production for many months now.  That it doesn't do something else -- *which you have yet to provide the requested reproduction details for* -- does not equal the patch not doing "what it should do".

bogdanpetrea:
@dalu: I've written a simple per resource authentication bypass: https://github.com/presslabs/keycloak-gatekeeper/pull/2 . It might be what you needed.

abstractj:
@pjeby could you please file a Jira (https://issues.jboss.org/browse/KEYCLOAK), add the affected version and steps to reproduce the issue? It helps us to plan/prioritize the next releases.

pjeby:
I've added [KEYCLOAK-8988](https://issues.jboss.org/projects/KEYCLOAK/issues/KEYCLOAK-8988) as per your request.

abstractj:
@pjeby thank you!

abstractj:
Thanks for spending some time on this. We really appreciate it. However I do not think this is something we want to add right now, because it's out of current scope for Gatekeeper, but in the future this can change. If you still think this is something valid to discuss, please do not hesitate to create a thread on keycloak-dev.
Thanks so much!

PR #348, opened 2018-04-20
Make use of go-oidc from the upstream

abstractj:
Would be nice if possible to use `go-oidc` from the upstream repository.
I can see several benefits:

* Avoid the burden keep up to date with the upstream
* Get the latest bug fixes from the commiters
* Prevent duplicated effort

I ran all the tests and they seem to be fine. Also, if there's anything which you would like to see in the upstream let me know, and I can try to talk with the authors and get these changes in.


abstractj:
It requires more changes, I'm closing it for now and will reopen after my update.

PR #350, opened 2018-04-23
Replace hash algorithm

abstractj:


gambol99:
much appreciated @abstractj  ... sorry for the delay

PR #353, opened 2018-05-04
Remove unused getUserInfo

abstractj:
@gambol99 I could not find any usage to `getUserinfo` inside the codebase. Do you think we can remove it?

### DIFF
--- a/oauth.go
+++ b/oauth.go
@@ -16,11 +16,6 @@ limitations under the License.
 package main
 
 import (
-	"encoding/json"
-	"errors"
-	"fmt"
-	"io/ioutil"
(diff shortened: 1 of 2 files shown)


abstractj:
@gambol99 I see. There's nothing in the spec which specifies that `UserInfo` endpoint should be only provided by the server, although Keycloak exposes it via `/realms/{realm-name}/protocol/openid-connect/userinfo`. The reason why I sent this was because such function is never exposed and have no usage.

Anyways, we don't need to take such decision right now. With the migration to the upstream from go-oidc we get this for free if we need.

I'm closing this and we can revisit in the future.

PR #357, opened 2018-05-12
Session Only Cookie

gambol99:
- adding a command line parameter to override the expiration on the cookie to make them session only


gambol99:
Add's an option to enable session only cookies as suggested in https://github.com/gambol99/keycloak-proxy/issues/343

PR #364, opened 2018-05-24
Add support for Strings claim

stefan-improbable:
This fixes https://github.com/gambol99/keycloak-proxy/issues/314


middleware.go
```diff
@@ -211,6 +212,60 @@ func (r *oauthProxy) authenticationMiddleware(resource *Resource) func(http.Hand
...
 
+// checkClaim checks whether claim in userContext matches claimName, match. It can be String or Strings claim.
+func (r *oauthProxy) checkClaim(user *userContext, claimName string, match *regexp.Regexp, resourceURL string) bool {
+	errFields := []zapcore.Field{
+		zap.String("claim", claimName),
+		zap.String("access", "denied"),
+		zap.String("email", user.email),
+		zap.String("resource", resourceURL),
+	}
+
```
gambol99: given [Claims](https://github.com/gambol99/go-oidc/blob/master/jose/claims.go#L10) is `type Claims map[string]interface{}` .. we could just and fail quick .. What do you think?

```
if _, found := user.claims[claimName]; !found {
    r.log.Warn("the token does not have the claim", errFields...)
    return false
}

stefan-improbable: Yep, I've added the statement here.

stefan-improbable:
@gambol99 I've added some info to README

gambol99:
cheers @stefan-improbable  ... LGTM

stefan-improbable:
@gambol99 thanks for the quick review!

PR #376, opened 2018-06-07
Logout Redirect Defaults

gambol99:
- change the default behaviour, if no redirect_ui set we default to the rediection-url and if not set we default to the Host header


gambol99:
@daniel-ac-martin

daniel-ac-martin:
Thanks @gambol99!

PR #377, opened 2018-06-08
Removal of HTTPClient

abstractj:
As part of our migration to coreos/go-oidc v2, I've been doing some
clean up. While doing this I found out that HTTPClient is never used
and seems to be not necessary anymore

@gambol99 this is part of our plans to move to the upstream. I'd appreciate your feedback.

### DIFF
--- a/server.go
+++ b/server.go
@@ -102,7 +102,7 @@ func newProxy(config *Config) (*oauthProxy, error) {
 
 	// initialize the openid client
 	if !config.SkipTokenVerification {
-		if svc.client, svc.idp, svc.idpClient, err = svc.newOpenIDClient(); err != nil {
+		if svc.client, svc.idp, err = svc.newOpenIDClient(); err != nil {
 			return nil, err
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -674,19 +674,18 @@ func (r *oauthProxy) newOpenIDClient() (*oidc.Client, oidc.ProviderConfig, *http
 			ID:     r.config.ClientID,
 			Secret: r.config.ClientSecret,
 		},
-		HTTPClient:        hc,
```
gambol99: so the client is no longer used by the underlining lib? .. as the probably break the https://github.com/gambol99/keycloak-proxy/blob/master/server.go#L628-L647 .. I recall this was added to enable skip-tls, custom proxy setting etc ..

abstractj: @gambol99 as far as I can tell, nothing was broken after the removal. But we can run some tests before take the decision of merging it.

sebastienblanc:
I thnik this is safe to merge, my tests passes, I have othre issues but not related to this PR

gambol99:
much appreciated @abstractj .. apologizes for the delay

Issue #331, opened 2018-03-20
Redirection URL doesn't work


waseem-h:
Hi, i've been using your keycloak-proxy for a while now and i'm facing a minor issue. Basically I have your proxy running as a sidecar container with an app on kubernetes. The issue is that the app itself is on someurl.com/app and whenever i visit the url, it takes me to keycloak, but the redirect_url in the browser url shows "'someurl.com" instead of someurl.com/app. So after logging in I am redirected to someurl.com which is weird

i've tried passing the argument `redirection-url=/app` but it doesn't work as well. Here's my config

```yaml
client-id: <client-id>
client-secret: <client-secret>
discovery-url: https://<keycloak-url>/auth/realms/<realm>
enable-default-deny: true
listen: 0.0.0.0:80
upstream-url: http://127.0.0.1:5601
redirection-url: /kibana
secure-cookie: false
verbose: true
enable-logging: true
cors-origins:
- '*'
cors-methods:
- GET
- POST
resources:
- uri: /

```

gambol99:
hi @waseem-h ... the redirection-url should the full url of the site ... i.e. https://example.com .. Assuming the redirection-url is not [provided it will use the Host header](https://github.com/gambol99/keycloak-proxy/blob/master/handlers.go#L39-L59)

Josh-West-Bose:
I'm running into this issue as well unfortunately, even with a fully qualified redirection URL.

I get sent to https://<host>/oauth/callback instead of https://<host>/<stuff>/oauth/authorize?...

waseem-h:
@gambol99 i tried with full url but still it doesn't seem to work

dlohin:
I am somewhat seeing a similiar issue with Kibana in a sidecar.  When I got to the URL, I am redirected to Keycloak (good) I logon and then I get sent to oauth/callback (as documented) but I am given access denied.

When I look at the Keycloak Proxy logs I see: 1.5296725145237534e+09	error	unable to exchange code for access token	{"error": "invalid_grant: Code not valid"}

After I have authenticated, though I can erase the oauth/callback uri and then hit enter again and it goes through fine.  Admittedly I don't entirely know what I am doing here, but is the oauth callback stripped by the proxy or does that make it back to the web service?  I am assuming I have some errors somwhere.

        args:
          - --listen=0.0.0.0:8080
          - --discovery-url=https://keycloak.local/auth/realms/edcop
          - --client-id=kibana
          - --client-secret=<removed>
[...]

PR #390, opened 2018-07-09
BaseURI parameter for deployments with non-root mount point

codepainters:
See issue #331 

This pull request is my quick fix for the problem listed there. Basically, I've added BaseURI parameter which is used to prefix all generated URIs, so it works mounted at the non-root URI of the upstream proxy.

### DIFF
--- a/config.go
+++ b/config.go
@@ -61,7 +61,11 @@ func newDefaultConfig() *Config {
 
 // WithOAuthURI returns the oauth uri
 func (r *Config) WithOAuthURI(uri string) string {
-	return fmt.Sprintf("%s/%s", r.OAuthURI, uri)
+	if r.BaseURI != "" {
+		return fmt.Sprintf("%s/%s/%s", r.BaseURI, r.OAuthURI, uri)
+	} else {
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -61,7 +61,11 @@ func newDefaultConfig() *Config {
 
 // WithOAuthURI returns the oauth uri
 func (r *Config) WithOAuthURI(uri string) string {
-	return fmt.Sprintf("%s/%s", r.OAuthURI, uri)
+	if r.BaseURI != "" {
```
gambol99: perhaps we can use the below rather than if statement

```Go
filepath.Join(r.BaseURI, r.OAuthURI, uri)
```

as it will handle a zero string anyhow ..

codepainters: I had it in mind while typing, but somehow felt a mismatch between URI and **file** path (read: backslashes on Windows) - yet I don't expect anyone to run this proxy on Windows, so I think it's fine.

gambol99: make's sense ... we should keep the `if` for now ..

gambol99:
LGTM ... much appreciated @codepainters

PR #392, opened 2018-07-10
Request ID

gambol99:
- adding a request id middleware injection via the --enable-request-id
- fixing up the Makefile

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -26,6 +26,7 @@ import (
 	"github.com/PuerkitoBio/purell"
 	"github.com/gambol99/go-oidc/jose"
 	"github.com/go-chi/chi/middleware"
+	uuid "github.com/satori/go.uuid"
 	"github.com/unrolled/secure"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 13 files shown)


middleware.go
```diff
@@ -66,6 +66,19 @@ func entrypointMiddleware(next http.Handler) http.Handler {
 	})
 }
 
+// requestIDMiddleware is responsible for adding a request id if none found
+func (r *oauthProxy) requestIDMiddleware(header string, length int) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			if v := req.Header.Get(header); v == "" {
+				req.Header.Set(header, randomString(length))
```
jangaraj: Why do we need `randomString`? IMHO better option is to use some UUID library - maybe github.com/google/uuid, also caddy uses it: https://github.com/mholt/caddy/blob/master/caddyhttp/requestid/requestid.go

gambol99:
```shell
[jest@starfury keycloak-proxy]$ make bench
--> Running go bench
goos: linux
goarch: amd64
pkg: github.com/gambol99/keycloak-proxy
BenchmarkRandomBytes36-8       	10000000	       146 ns/op	      96 B/op	       2 allocs/op
BenchmarkRandomString36-8      	10000000	       148 ns/op	      96 B/op	       2 allocs/op
# using google uuid 
BenchmarkUUID-8                	 1000000	      1183 ns/op	      16 B/op	       1 allocs/op
# using a random string 
BenchmarkRandomUUID-8          	10000000	       187 ns/op	      80 B/op	       2 allocs/op
BenchmarkEncryptDataBlock-8    	 1000000	      2303 ns/op	    1200 B/op	       7 allocs/op
BenchmarkEncodeText-8          	  500000	      3262 ns/op	    2352 B/op	      11 allocs/op
BenchmarkDecodeText-8          	 1000000	      1907 ns/op	    2208 B/op	      10 allocs/op
BenchmarkContainsSubString-8   	200000000	         7.95 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/gambol99/keycloak-proxy	17.928s

```
Running both with a 36 character length, it's 10x slower

gambol99: basically i got bored and started messing about and saved 1000ns :-)

jangaraj: 👍 Could you try https://github.com/edwingeng/wuid, pls?

PR #405, opened 2018-07-27
Add connection pool settings for performance tuning.

jgroffen:
This pull request includes two new settings that allow users to tune the connection pool. This pool manages connections to keycloak and the upstream server.

These settings allowed me to tune keycloak-proxy to ~6500 requests a second, sustained for 20 minutes, at which point I was limited by ethernet bandwitdh.

Default settings for net.http are 100 max idle connections, but only 2 max idle connections per host. The net.http library will create and destroy a connection without taking advantage of keepalive when the idle connection pool runs out. Under heavy load this leads to many ephemeral ports in TIME_WAIT state, and the server eventually runs out of ephemeral ports while waiting for these ports to free up. At this point the server will start dropping requests.

Keep in mind that keycloak-proxy only really talks to the keycloak server and the upstream server, so there are only 2 hosts involved, meaning only 4 connections in the pool will be utilised. Because of this I changed the default max-idle-connections-per-host to 50.

Our situation involves spikes of high load - thousands of requests in short bursts, so we needed to be able to tune the connection pool.

An article I found helpful that describes how idle connections work in net.http: http://tleyden.github.io/blog/2016/11/21/tuning-the-go-http-client-library-for-load-testing/

The implementation in this pull request is minimal intentionally, given the pending move to the keycloak repo. I was initially thinking it may be better to manage a separate connection pool for the upstream server and the keycloak server (and any other servers the proxy may communicate with). I haven't come across any other discussion of load problems with keycloak-proxy, so I think being able to tune the shared connection pool (implemented in this pull request) will likely meet the need long term.

If you have any feedback or suggestions, please let me know.

Thanks!


jgroffen:
Updated readme to reference the correct pull request

jgroffen:
Please note: adding validation rules for the two new settings caused quite a number of changes to config_test.go, though the changes are simple. I tried to retain the intent of the test cases, and add new tests for the config validation rules.

gambol99:
much appreciated @jgroffen ... LGTM!

PR #407, opened 2018-07-30
Migration to go-oidc from upstream v1

abstractj:
This change aims to make use of coreos/go-oidc. The generic adapter
aims to act as a sidecar, provide a way to SkipClientID is out of scope
for now.

That does not mean that we cannot revisit this in the near future.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -28,6 +28,18 @@
   revision = "144418e1475d8bf7abbdc48583500f1a20c62ea7"
 
 [[projects]]
+  branch = "v1"
+  name = "github.com/coreos/go-oidc"
+  packages = [
+    "http",
+    "jose",
+    "key",
+    "oauth2",
+    "oidc"
+  ]
+  revision = "e860bd55bfa7d7cb35d30d26a167982584f616b0"
+
+[[projects]]
   name = "github.com/coreos/pkg"
   packages = [
     "health",
@@ -42,28 +54,16 @@
   revision = "5215b55f46b2b919f50a1df0eaa5886afe4e3b3d"
 
 [[projects]]
+  name = "github.com/elazarl/goproxy"
+  packages = ["."]
+  revision = "947c36da3153ff334e74d9d980de341d25f358ba"
+  version = "v1.1"
+
+[[projects]]
   name = "github.com/fsnotify/fsnotify"
   packages = ["."]
   revision = "c2828203cd70a50dcccfb2761f8b1f8ceef9a8e9"
   version = "v1.4.7"
-
-[[projects]]
-  branch = "master"
-  name = "github.com/gambol99/go-oidc"
-  packages = [
-    "http",
-    "jose",
-    "key",
-    "oauth2",
-    "oidc"
-  ]
-  revision = "87948fe50989333a6a3f970e48f8dec844361fa1"
-
-[[projects]]
-  branch = "master"
-  name = "github.com/gambol99/goproxy"
-  packages = ["."]
-  revision = "e713e5909438245be49ef559e74dd904833ebe90"
 
 [[projects]]
   name = "github.com/go-chi/chi"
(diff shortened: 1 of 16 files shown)


jangaraj:
Why github.com/coreos/go-oidc v1 and not v2? v1 seems to be a deprecated branch.

Gopkg.lock
```diff
@@ -27,6 +27,18 @@
   packages = ["."]
   revision = "144418e1475d8bf7abbdc48583500f1a20c62ea7"
 
+[[projects]]
+  branch = "v1"
```
slaskawi: I agree, we should probably use `v2` if possible. Otherwise LGTM.

abstractj:
@jangaraj @slaskawi I understand your concern and it's a fair argument. Although, v1 and v2 are completely different APIs and there's a huge break of API compatibility. For example:

- coreos/go-oidc/oauth2 turns into golang/x/oauth2
- coreos/go-oidc/jose turns into square/go-jose
- coreos/go-oidc/oidc turns into coreos/go-oidc

And they are completely different APIs, which requires a considerable amount of time migrating the codebase. Actually, we already have been working on it here: https://github.com/gambol99/keycloak-proxy/commits/develop.

This proposal to target `v1` does not mean that things will be set in stone. But this change allows us to move forward and migrate this repository to Keycloak organization, and later revisit and continue the work to upgrade to `v2`.

magnock:
Hello,

Something changed since the merge, I've been using latest docker image with Azure AD Oauth2, it was working fine till 2 days ago.
Now I'm having this error : unable to verify the id token	{"error": "oidc: JWT claims invalid: Invalid claim value: 'iss'. expected=https://login.microsoftonline.com/xxxxxxxxxxxxx/v2.0, found=https://sts.windows.net/xxxxxxxxxxxx/"}

It looks like the new merged oidc makes a request with no version, and receives a v1 Azure AD response (sts.windows.net/xxxxxxx) which is not expected. 

I've tried to move to an old docker image, but I have the same problem. which I did'nt understand !!!

Thanks for your help.

abstractj:
@magnock I'm sorry but I'm not sure how this pull request relates with what you mentioned. Nothing was merged into master yet.

gambol99:
hi @magnock  ... perhaps something has changed upstream in Azure as I can't see anything that would have effected this between this and the last version [here](https://github.com/gambol99/keycloak-proxy/compare/v2.2.2...v2.3.0) .. At the error is being thrown in the go-oidc lib which hasn't changed on master ... I recall their used to be an exception in go-oidc for google .. which used to do something similar, i.e. being issued a token outside of the issuer you called ...

i did see https://github.com/coreos/go-oidc/issues/121 .. I've not read it fully ..

jangaraj:
BTW: master build is failing - https://travis-ci.org/gambol99/keycloak-proxy/builds/420095894?utm_source=github_status&utm_medium=notification

abstractj:
@jangaraj thanks for reporting this. I will take a look at what's going on.

lukasmrtvy:
@magnock Did you figure out, where is a problem?

abstractj:
@lukasmrtvy would you mind to file a Jira (https://issues.jboss.org/projects/KEYCLOAK) with a proper description, steps to reproduce and affected version? We moved from GH issues, and important things like this can be easily missed if commented on closed PRs.

Thanks in advance.

marrobi:
Apologies for resurfacing an old issue, however I'm getting the same error as @magnock @lukasmrtvy above with Azure AD:

```
  error   unable to verify the id token   {"error": "oidc: JWT claims invalid: invalid claim value: 'iss'. expected=https://login.microsoftonline.com/xxxxx/v2.0, found=https://sts.windows.net/xxxxx/."}
```

Despite a v2 endpoint, a v1 token seems to be returned.  Did anyone find a solution? 

I can create a new issue if feel that would be useful. Thanks.

Issue #409, opened 2018-08-15
Large kc_access cookies appear to be improperly decoded prior to parsing


jcrowthe:
I am looking to utilize keycloak-proxy with Azure AD using OIDC. It appears to be functional, with the expected JWT token being return to the keycloak-proxy, but I receive the following errors in redirect loop:

```
...
error	keycloak-proxy/middleware.go:108	no session found in request, redirecting for authorization	{"error": "malformed JWS, only 2 segments"}
debug	keycloak-proxy/handlers.go:81	incoming authorization request from client address	{"access_type": "", "auth_url": "<my-url>", "client_ip": "10.1.7.66:38966"}
warn	keycloak-proxy/handlers.go:140	unable to parse the access token, using id token only	{"error": "malformed JWS, only 1 segments"}
info	keycloak-proxy/handlers.go:160	issuing access token for user	{"email": "", "expires": "2018-08-15T20:34:58Z", "duration": "59m59.838099673s"}
error	keycloak-proxy/middleware.go:108	no session found in request, redirecting for authorization	{"error": "malformed JWS, only 2 segments"}
debug	keycloak-proxy/handlers.go:81	incoming authorization request from client address	{"access_type": "", "auth_url": "<my-url>", "client_ip": "10.1.7.66:38966"}
warn	keycloak-proxy/handlers.go:140	unable to parse the access token, using id token only	{"error": "malformed JWS, only 1 segments"}
info	keycloak-proxy/handlers.go:160	issuing access token for user	{"email": "", "expires": "2018-08-15T20:34:58Z", "duration": "59m59.037853008s"}
error	keycloak-proxy/middleware.go:108	no session found in request, redirecting for authorization	{"error": "malformed JWS, only 2 segments"}
debug	keycloak-proxy/handlers.go:81	incoming authorization request from client address	{"access_type": "", "auth_url": "<my-url>", "client_ip": "10.1.7.66:38966"}
warn	keycloak-proxy/handlers.go:140	unable to parse the access token, using id token only	{"error": "malformed JWS, only 1 segments"}
...
```

In taking a packet capture of keycloak-proxy (SSL offloaded previously), I receive the following (JWT obscured):

```
HTTP/1.1 307 Temporary Redirect
Content-Type: text/html; charset=utf-8
Location: /
Set-Cookie: kc-access=eyJ0eXAi...; Path=/; Domain=dashboard.k8s-alpha.ethos.adobe.net; HttpOnly
[...]

PR #410, opened 2018-08-15
Reduce cookie divide limit to 4000 bytes

jcrowthe:
Cookies have a size limit of 4096 bytes. Currently, cookies in this project are divided at 4089 bytes to accommodate, however it does not take into account the Path, Domain, and Expiration attributes of a cookie, which also count toward the limit.

This PR decreases the cookie size limit to 4000 bytes, which should account for most use cases (barring inordinately long DNS names). See issue #409 for more details.


jangaraj:
This is only a workaround for your use case. It will be an issue again for other users with very long domain name. It really needs more precise `maxCookieLength` calculation in the code. Metacode:
```
maxCookieLength := 4089 - len(r.config.CookieAccessName) - sizeOfOthersCookieElements(domainsize, pathsize, datesize, ...)
```

jcrowthe:
I'll accept that. Feel free to close this and submit a subsequent PR to address this more precisely.

jangaraj:
I'm only 3rd party contributor, so I can't close PR. Sure, I'll create my PR.

stianst:
Closing the PR as requested

PR #419, opened 2018-09-17
[KEYCLOAK-9033] Add base URI to OAuth routes

pgagnon:
This PR implements a simple fix that ensures that keycloak-proxy prefixes the required oauth endpoints with the value defined in the `base-uri` option.

This is required in order to work, for instance, in environments where keycloak-proxy is used behind another reverse proxy that directs traffic to different backends based on the path requested.

### DIFF
--- a/cookies.go
+++ b/cookies.go
@@ -32,11 +32,15 @@ func (r *oauthProxy) dropCookie(w http.ResponseWriter, host, name, value string,
 	if r.config.CookieDomain != "" {
 		domain = r.config.CookieDomain
 	}
+	path := r.config.BaseURI
+	if path == "" {
+		path = "/"
+	}
 	cookie := &http.Cookie{
 		Domain:   domain,
 		HttpOnly: r.config.HTTPOnlyCookie,
 		Name:     name,
-		Path:     "/",
+		Path:     path,
 		Secure:   r.config.SecureCookie,
 		Value:    value,
 	}
--- a/cookies_test.go
+++ b/cookies_test.go
@@ -40,6 +40,45 @@ func TestCookieDomainHostHeader(t *testing.T) {
 	assert.Equal(t, cookie.Domain, "127.0.0.1")
 }
 
+func TestCookieBasePath(t *testing.T) {
+	cfg := newFakeKeycloakConfig()
+	cfg.BaseURI = "/base-uri"
+
+	_, _, svc := newTestProxyService(cfg)
+
+	resp, err := makeTestCodeFlowLogin(svc + "/admin")
+	assert.NoError(t, err)
+	assert.NotNil(t, resp)
+
+	var cookie *http.Cookie
+	for _, c := range resp.Cookies() {
+		if c.Name == "kc-access" {
+			cookie = c
+		}
+	}
+	assert.NotNil(t, cookie)
+	assert.Equal(t, "/base-uri", cookie.Path)
+}
+
+func TestCookieWithoutBasePath(t *testing.T) {
+	cfg := newFakeKeycloakConfig()
+
+	_, _, svc := newTestProxyService(cfg)
+
+	resp, err := makeTestCodeFlowLogin(svc + "/admin")
+	assert.NoError(t, err)
+	assert.NotNil(t, resp)
+
+	var cookie *http.Cookie
+	for _, c := range resp.Cookies() {
+		if c.Name == "kc-access" {
+			cookie = c
+		}
+	}
+	assert.NotNil(t, cookie)
+	assert.Equal(t, "/", cookie.Path)
+}
+
 func TestCookieDomain(t *testing.T) {
 	p, _, svc := newTestProxyService(nil)
 	p.config.CookieDomain = "domain.com"
--- a/handlers.go
+++ b/handlers.go
@@ -209,10 +209,6 @@ func (r *oauthProxy) oauthCallbackHandler(w http.ResponseWriter, req *http.Reque
 			redirectURI = string(decoded)
 		}
 	}
-	if r.config.BaseURI != "" {
-		// assuming state starts with slash
-		redirectURI = r.config.BaseURI + redirectURI
-	}
 
 	r.redirectToURL(redirectURI, w, req, http.StatusTemporaryRedirect)
 }
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -353,6 +353,44 @@ func TestOauthRequests(t *testing.T) {
 	newFakeProxy(cfg).RunTests(t, requests)
 }
 
+func TestOauthRequestsWithBaseURI(t *testing.T) {
+	cfg := newFakeKeycloakConfig()
+	cfg.BaseURI = "/base-uri"
+	requests := []fakeRequest{
+		{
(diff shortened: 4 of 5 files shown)


pgagnon:
@gambol99 Did anything change for getting this merged due to the move to keycloak org?

abstractj:
@pgagnon could you please file a Jira (https://issues.jboss.org/browse/KEYCLOAK), add the affected version and steps to reproduce the issue? It helps us to plan/prioritize the next releases.

pgagnon:
@abstractj Please see [KEYCLOAK-9033](https://issues.jboss.org/browse/KEYCLOAK-9033). Thank you.

FrodeI:
I think this commit will break configurations that rewrites the uri in the reverse proxy.
Ex:
 - External URL for application: https://<domain>/myapp/<some-path>
 - Ingress terminates SSL and rewrites this to: http://<gatekeeper-service>/<some-path>
 - Gatekeeper is running with:
    --base-uri=/myapp
    --redirection-url=https://<domain>/

It might be possible to use ingress without path rewrite by setting "--oauth-uri /myapp/oauth" instead, but I haven't tested it.

pgagnon:
@Frodel Unless I misunderstand, we are running the configuration you describe with this commit in production without issue.

wodansax:
I can confirm, that this change fixes the problem when used with path-based loadbalancing.

dlisin:
I think `base-url` option should also be applied to access/refresh cookies set by `gatekeeper`
Currently these cookies are set as domain wide. So if you are running multiple gatekeepers on same domain, but with different base URL's - they will interfere with each other

pgagnon:
@dlisin I agree with this. This is a necessary improvement in order to make keycloak-gatekeeper viable in a path-based routing scenario.

dovreshef:
What's the status of this PR? Anything I can do to help?
I'm also running it behind a path prefix.

pgagnon:
@dlisin It looks like adding the base path to cookies is rather trivial. I am doing some internal testing and will update my branch with your proposed improvement. Thanks!

pgagnon:
@dovreshef At this point I think we need a formal review from the maintainers. @abstractj Is there anything I can do to get the process started?

wodansax:
Any update on this?

pgagnon:
@abstractj I'll find some time to add a couple of tests tomorrow or over the weekend. 👍

pgagnon:
@abstractj Added tests for cookies path and oauth base uri.

dlisin:
Dear Team, any chance to have this merged?

abstractj:
@dlisin I can promise to give a review, if that makes sense.

dlisin:
@abstractj it will be great! This PR is only missing think, that blocks us to move to keycloak-gatekeeper from our own fork

berkitamas:
I saw that you added the r.config.BaseURI behind the slash in the commit d75ee83. Your modification will be invalid on http://\<domain\>/\<base-uri\> since cookies with Path: /path/ property are not available on /path.

pgagnon:
@berkitamas I will adjust ASAP. Thanks for catching this.

abstractj:
@pgagnon I would like to make more tests, just to be safe. Meanwhile, could you please squash the 6 commits in a single one?

pgagnon:
@abstractj Would you like me to extract the changes related to cookies path in a second PR, or just squash everything and proceed with only this one?

abstractj:
@pgagnon just squash everything should be enough. Thanks Philippe.

pgagnon:
@abstractj Squashed + added a test for cookies path without base path value.

pgagnon:
@abstractj Friendly ping.

pgagnon:
Thanks, @abstractj !

abstractj:
Thank you @pgagnon!

mboret:
@pgagnon How did you configure the custom base-uri? Because I've my service behind an nginx ingress controller, "/service1" with the keycloak gatekeeper as a sidecar. When I set the parameter "base-uri" to "/service1", this creates a redirection loop.
Keycloak log:

>  "msg: no session found in request, redirecting for authorization ,error: authentication session not found".

I've tried to add the parameter "redirection-url" with the value "https://app.example.com/service1" but the same issue.

pgagnon:
@mboret Would you be able to share your nginx configuration? There may be a case I have overlooked. If there is I will prepare a fix PR.

mboret:
@pgagnon For sure. 

My Nginx configuration is standard, I mean deployed with the pre-defined stable/nginx Helm chart configuration.

Kubernetes Ingress Configuration(I tried to remove "ingress.kubernetes.io/rewrite-target: /" but same):

    apiVersion: extensions/v1beta1
    kind: Ingress
    metadata:
    annotations:
        ingress.kubernetes.io/rewrite-target: /
        kubernetes.io/ingress.class: nginx
        nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    labels:
        app: service1
        component: server
        heritage: Tiller
    name: service1-server
    namespace: test-service1
    spec:
    rules:
    - host: test.example.com
        http:
        paths:
        - backend:
            serviceName: service1-server
            servicePort: 80
            path: /service1
    tls:
    - hosts:
        - test.example.com
        secretName: cert-test.example.com
    status:
    loadBalancer:
        ingress:
        - {}

service1 Deployment Configuration:

    apiVersion: extensions/v1beta1
    kind: Deployment
    metadata:
    name: service1-deployment
    spec:
    replicas: 1
    template:
        metadata:
        name: service1
        spec:
        containers:
        - name: service1
            image: grc.io/service1:latest
            args:
            - --listen-port=8080
            - --env=test
        - name: proxy
            image:  keycloak/keycloak-gatekeeper:5.0.0
            imagePullPolicy: Always
            args:
            - --resource=uri=/*
            - --discovery-url=https://auth.example.com/auth/realms/MYREALM
            - --client-id=service1
            - --client-secret=XXXXXXXX
            - --listen=0.0.0.0:3000
            - --enable-logging=true
            - --enable-json-logging=true
            - --base-uri=/service1
            - --upstream-url=http://127.0.0.1:8080

wodansax:
Have you tried this (without service1):
`- --redirection-url=http://test.example.com`

### Addtionally:
#### Is the upstream-url correct?
I don't use multiple containers in one Deployment by myself, but the 127.0.0.1 seems a bit odd, since they are 2 separate containers?

#### Is the service configured correctly?
He should point from port 80 to port 3000 i guess?

mboret:
@wodansax Yes, I'd already tried this. It seems this "redirection-url" parameter do nothing because I tried with "https//google.com", and same. The only difference is with "http", this generates a TLS error configure.

Containers in the same pod share the network namespace. 
And if the service I configured correctly. Without the keycloak sidecar, service1, works.

Josh-West-Bose:
@mboret I'm having similar issues when behind an Ingress Controller that also does PathPrefixStrip (or rewrite-target style) operations

abstractj:
@mboret @Josh-West-Bose would you mind to file a [Jira](https://www.keycloak.org/community.html) for that adding all the details, steps to reproduce and affected version (the more detailed, the better) ? In this way we can evaluate and look at the issue. Otherwise, we can miss it.

PR #432, opened 2018-11-02
Expand resource validation help text to match syntax

emsearcy:
- Update resource syntax to include all available identifiers.
- Reuse the same string for both parsing errors and when the identifier doesn't match.
- Fix grammar it's -> its (possessive form).

### DIFF
--- a/resource.go
+++ b/resource.go
@@ -33,10 +33,11 @@ func (r *Resource) parse(resource string) (*Resource, error) {
 	if resource == "" {
 		return nil, errors.New("the resource has no options")
 	}
+	resourceHelp := "invalid resource keypair, should be (uri|roles|groups|methods|white-listed|require-any-role)=comma_values"
 	for _, x := range strings.Split(resource, "|") {
 		kp := strings.Split(x, "=")
 		if len(kp) != 2 {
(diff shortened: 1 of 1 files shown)


abstractj:
@emsearcy could you please file a Jira (https://issues.jboss.org/browse/KEYCLOAK), add the affected version and steps to reproduce the issue? It helps us to plan/prioritize the next releases.

emsearcy:
Hi - I didn't see the notification from your earlier comment.  These are simple text-only improvements to the help text to align your help text with actual behavior to help your users who are trying to figure out how your software works without digging into to the codebase to find out, like I had to do.  I'm sorry, but I've also had poor developer engagement on what I have opened in JBOSS, so if you don't want to accept small, useful contributions at face value, I think that is to the detriment of your community and users.

abstractj:
@emsearcy np and like I mentioned, feel free to reopen. 

I understand your frustration. But we use Jiras for absolutely everything, documentation, bugs, feature requests and etc. In a such big project, that's the way how the releases are planned.

Also, If there's anything we can do better to make your developer engagement experience better, feel free to send a feedback on keycloak-dev mailing list.

PR #433, opened 2018-11-12
OAuth2 state parameter

stianst:
- [KEYCLOAK-8669] Gatekeeper state parameter generated doesn’t have
enough entropy
- [KEYCLOAK-8667] CSRF vulnerability in Gatekeeper due to the lack of
checks for "state" parameter


FrodeI:
This merge gives a warning in the authorization callback which assumes the state should be b64-encoded:
warn	unable to decode the state parameter	{"state": "e75defde-dde6-47d5-894d-3cad6fd3bb47", "error": "illegal base64 data at input byte 8"}

abstractj:
@FrodeI please file a Jira reporting the issue, affected version and steps to reproduce. Otherwise, your comment can be missed. Thanks in advance.

PR #436, opened 2018-11-20
Add option to select oauth auth method

apinnecke:
Hi,

**TL;DR:**
This PR implements a config option to choose between basic and post authentication during communication between OAuth client and OAuth server. :v:

**Background**:
I am currently building a kubernetes infrastructure for one of my clients and stumbled across this issue yesterday during implementing this gatekeeper as an OIDC proxy for https://onelogin.com .  The issue I ran into is the following: https://stackoverflow.com/questions/52635508/what-is-the-meaning-of-this-error-client-authentication-must-only-be-provided , which states to use either basic *or* post authentication. 🤔

Due to the fact that the gatekeeper is still using go-oidc v1 (as decided and argued about in #407) it is still having a bug as described in coreos/go-oidc#181 which sets the post authentication even when basic is used, so the only quick chance for me to get this one up and running is to implement a config option to let the user choose between basic and post auth (which fixes the problem for me, tested with custom build in my cluster). 😑

### DIFF
--- a/config.go
+++ b/config.go
@@ -49,6 +49,7 @@ func newDefaultConfig() *Config {
 		MaxIdleConns:                100,
 		MaxIdleConnsPerHost:         50,
 		OAuthURI:                    "/oauth",
+		OAuthAuthMethod:             "basic",
 		OpenIDProviderTimeout:       30 * time.Second,
 		PreserveHost:                false,
 		SelfSignedTLSExpiration:     3 * time.Hour,
@@ -62,6 +63,7 @@ func newDefaultConfig() *Config {
 		SkipOpenIDProviderTLSVerify: false,
 		SkipUpstreamTLSVerify:       true,
 		Tags: make(map[string]string, 0),
+
 		UpstreamExpectContinueTimeout: 10 * time.Second,
 		UpstreamKeepaliveTimeout:      10 * time.Second,
 		UpstreamKeepalives:            true,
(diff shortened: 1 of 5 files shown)


abstractj:
@apinnecke could you please file a Jira (https://issues.jboss.org/browse/KEYCLOAK-8856), add the affected version and steps to reproduce the issue? It helps us to plan/prioritize the next releases.

apinnecke:
@abstractj this is a PR, not an issue. And you obviously didn‘t even read the PR?!

abstractj:
@apinnecke Jiras are not only for bugs, but for feature, tasks and other things as well. Please, read https://github.com/keycloak/keycloak-gatekeeper#help-and-documentation.

config.go
```diff
@@ -49,6 +49,7 @@ func newDefaultConfig() *Config {
 		MaxIdleConns:                100,
 		MaxIdleConnsPerHost:         50,
 		OAuthURI:                    "/oauth",
+		OAuthAuthMethod:             "basic",
```
stianst: Rename to 'ClientAuthMethod'. Supported values should be 'secret-basic' and 'secret-body'. It should be prefixed with 'secret-' to differentiate between other authentication mechanisms that may be added in the future. Also, it should be 'body' not 'post' (both basic and body are after all sent in a post).

abstractj:
It has been a while since we requested changes to this PR. While the submission is appreciated, we are not able to merge it without the changes request and a Jira as well. Because it has been so long, we are going to close the issue. Feel free to reopen with the changes included, if your time permits.

PR #437, opened 2018-11-20
Use the correct URL parameter for token revocation when refresh tokens disabled

GJKrupa:
In its current state, this logs a 400 error when trying to revoke the session from Keycloak if refresh tokens aren't enabled because it's trying to pass an ID token on the refresh_token parameter.  This change selects the correct parameter name based on which token is being used.

Also, ran "make format" to allow the gofmt check to pass.


abstractj:
@GJKrupa @robbiemcmichael could you please file a Jira (https://issues.jboss.org/browse/KEYCLOAK-8856), add the affected version and steps to reproduce the issue? It helps us to plan/prioritize the next releases.

GJKrupa:
Thanks.  I've raised https://issues.jboss.org/browse/KEYCLOAK-8990.

abstractj:
@GJKrupa thank you!

abstractj:
Hi @GJKrupa, could you please rebase and remove the commit 83f9fb4? I managed to manually reproduce the issue, but do you think you can attack some test cases for this scenario?

abstractj:
@GJKrupa  It has been a while since we requested changes to this PR. While the submission is appreciated, we are not able to merge it without these changes. Because it has been so long, we are going to close the issue. Feel free to reopen with the changes included, if your time permits.

GJKrupa:
Apologies that I haven't had time for this due to workload.  Probably best at this point to re-create it anyway rather than try to rebase after so long.

abstractj:
@GJKrupa absolutely no problem. I agree with you, but before you move forward with coding, let's figure out the exact steps to reproduce the issue.

PR #445, opened 2018-12-12
[KEYCLOAK-9074] Optionally enforce token encryption in cookie

fredbi:
* add config option to distinguish when to encrypt token: in header or/and in cookie
* fixes #9704

Point mentioned by enhancement proposal #9074.


abstractj:
@fredbi finally landed, thank you for contributing to this.

PR #447, opened 2019-01-01
[KEYCLOAK-9128] Vendor update as of 2019-01-01

fredbi:


fredbi:
@abstractj the vendor update caused tests to fail - the test update just reflects the change in the updated dependencies.

go.mod
```diff
@@ -0,0 +1,40 @@
+module github.com/fredbi/keycloak-gatekeeper
```
abstractj: @fredbi could you please replace your username by "keycloak"? Other than that we're good to go.

fredbi: @abstractj ah yes, that's this automatic go mod update... Fixing it thanks

fredbi:
@abstract I am removing the go.mod specification for the moment, as it is still difficult for CI to figure out some other broke dependencies. There are some other dependency fixes to be carried out to be able to properly build with go modules. I'd rather deferring this to another PR..

abstractj:
Superseded by https://github.com/keycloak/keycloak-gatekeeper/pull/466 which is just a rebase of this PR.

Thank you.

PR #448, opened 2019-01-01
[KEYCLOAK-9086] Linting

fredbi:
* abide by modern go linting standards (golangci)
* include `.golangci.yml` config to (optionally) carry out automated linting check on PRs
* does not change behavior

NOTE: upgrading deprecated prometheus middleware is not carried out in this PR


fredbi:
There is indeed a _suggestion_ to _add_ golangci as a CI specialized on the linting job.
At any rate, this is opt-in.

This is by no means a "forced push" to any given CI product and this is definitely not a proposal to replace TravisCI (basically, this would add another level of checks on your PRs, no more, no less).

Anyway, the presence of the .golangci.yml is not coupled with the use of the golangci CI: it is also a helper config for developers who, like me, are using this tool locally on the dev environment.

So basically, I may remove the .golangci.yml dot-file if you feel uneasy about this. If this is what you want, you just tell me.

However, I suggest you try their command line tool `golangci-lint` locally so you can see how it goes.

fredbi:
ok see #454

doc.go
```diff
@@ -62,6 +62,17 @@ const (
 	claimResourceAccess = "resource_access"
 	claimResourceRoles  = "roles"
 	claimGroups         = "groups"
+
+	accessCookie       = "kc-access"
+	refreshCookie      = "kc-state"
+	requestURICookie   = "request_uri"
+	requestStateCookie = "OAuth_Token_Request_State"
```
abstractj: @fredbi Even though I understand you would like to bring awareness. I'd suggest to just rename to `httpScheme` and `httpsScheme`.

handlers.go
```diff
@@ -55,9 +56,9 @@ func (r *oauthProxy) getRedirectionURL(w http.ResponseWriter, req *http.Request)
 		redirect = r.config.RedirectionURL
 	}
 
-	state, _ := req.Cookie("OAuth_Token_Request_State")
+	state, _ := req.Cookie(requestStateCookie)
 	if state != nil && req.URL.Query().Get("state") != state.Value {
-		r.log.Error("State parameter mismatch")
+		r.log.Error("state parameter mismatch")
```
abstractj: Out of curiosity, why you removed the capital letter at the start of each error message? Any specific reason? If there's no reason, my 2 cents is to leave it as is.

handlers.go
```diff
@@ -422,18 +423,19 @@ func (r *oauthProxy) tokenHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(user.token.Payload)
+	_, _ = w.Write(user.token.Payload)
```
abstractj: What's the difference between `w.Write(user.token.Payload)` and `_, _ = w.Write(user.token.Payload)` ?

abstractj:
@fredbi thank you for the refactoring. I added few comments and questions.

handlers.go
fredbi: this is one of these linting rules

doc.go
fredbi: httpsScheme stutters...

handlers.go
fredbi: Again, linting rule for code smells: one should not ignore returned values, unless doing so explicitly. Hence the _, _ to explicitly drop the result.

fredbi:
@abstractj should really should give a try to golangci-lint locally. At first, this looks pecky and intrusive with your code style, but given some time, you get used to all these conventions, and code smells are more easily detected.

doc.go
abstractj: @fredbi ok, that's fair enough.

abstractj:
@fredbi actually I did it, that's why I asked :) Locally here, it didn't complain about capitals or assignments. But probably is some misconfiguration on my end. Thanks for your submission.

PR #452, opened 2019-01-30
Add ability to proxy to an upstream sub path

bashofmann:
This allows to configure an upstream url with a path like `http://example.com/foo` and requests to `/bar` will be proxied to `http://example.com/foo/bar`.


abstractj:
@bashofmann thank you for your suggestion. I would appreciate if you write an e-mail to keycloak-dev with your suggestion. After that, please file a Jira, include tests for it and we will gladly review this PR.

abstractj:
@bashofmann it has been a while since we requested changes to this PR. While the submission is appreciated, we are not able to merge it without these changes. Because it has been so long, we are going to close the issue. Feel free to reopen with the changes included, if your time permits.

PR #456, opened 2019-02-10
Keep the response headers already set to the response before the upstream request

BertrandFritsch:
Enable the KeepDestinationHeaders flag of the upstream proxy to preserve the headers already set in the response before the request to the upstream server.

Concerned headers are the set-cookie header holding the refreshed access_token and the headers set with the --response-headers flag

### DIFF
--- a/server.go
+++ b/server.go
@@ -157,6 +157,8 @@ func (r *oauthProxy) createReverseProxy() error {
 	if err := r.createUpstreamProxy(r.endpoint); err != nil {
 		return err
 	}
+	// keep the destination headers to eventually preserve the refreshed access token cookie in the response
+	r.upstream.(*goproxy.ProxyHttpServer).KeepDestinationHeaders = true
 	engine := chi.NewRouter()
 	engine.MethodNotAllowed(emptyHandler)
 	engine.NotFound(emptyHandler)
--- a/server_test.go
+++ b/server_test.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	"github.com/coreos/go-oidc/jose"
+	"github.com/elazarl/goproxy"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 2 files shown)


BertrandFritsch:
Open issue: https://issues.jboss.org/browse/KEYCLOAK-9528

fredbi:
Interesting. Your fix corresponds exactly to my first commit for #441.
I stumbled upon this with cors headers but i understand now that this is a more general issue.
Also your report now justifies this setting as mandatory and not optional (as was requested for my #441 durIng review).

fredbi:
Also do you have some unit test on this?
Fyi i added some e2e simulation in #441 but obviously, did not cover this specific case with expiration.

BertrandFritsch:
Actually no I don't have any other unit test than the existing one that I've completed to check that the KeepDestinationHeaders flag is set to true on the goproxy upstream request. I don't see any other way to test it as the expected behavior is implemented by a third-party library.

abstractj:
@fredbi @BertrandFritsch would make sense to get this PR incorporated into https://github.com/keycloak/keycloak-gatekeeper/pull/441?

BertrandFritsch:
@abstractj No problem for me if it can speed up the process of publishing a new version with this fix. I let @fredbi decide which would be the best choice

fredbi:
I think it is ok to onboard this with #441. 
@abstractj how about the optional set up you asked for? shall we make it systematic now?

fredbi:
Onboarded on #441 which I rebased. This reports shows that this setup should not be optional.

abstractj:
@BertrandFritsch thank you I'm closing this because this PR is superseded by https://github.com/keycloak/keycloak-gatekeeper/pull/441. But feel free to discuss there, or if there's anything different than that to address. Just open a new PR.

PR #463, opened 2019-03-09
[KEYCLOAK-9786] Secure token and logout endpoint

fredbi:
* now token validity is checked to reach those endpoints, even though a valid cookie is presented
* previous BadRequest responses on malformed tokens now yield Unauthorized


abstractj:
@fredbi thanks for contributing to this.

PR #469, opened 2019-03-17
Fix function comments based on best practices from Effective Go

CodeLingoTeam:
Hi, we updated an exported function comment based on best practices from [Effective Go](https://golang.org/doc/effective_go.html). It’s admittedly a relatively minor fix up. Does this help you?

### DIFF
--- a/stores.go
+++ b/stores.go
@@ -54,7 +54,7 @@ func (r *oauthProxy) StoreRefreshToken(token jose.JWT, value string) error {
 	return r.store.Set(getHashKey(&token), value)
 }
 
-// Get retrieves a token from the store, the key we are using here is the access token
+// GetRefreshToken gets retrieves a token from the store, the key we are using here is the access token
 func (r *oauthProxy) GetRefreshToken(token jose.JWT) (string, error) {
 	// step: the key is the access token
 	v, err := r.store.Get(getHashKey(&token))


abstractj:
@CodeLingoBot thanks for your contribution. But I'd rather to have a full review of all comments in the codebase. So we can provide a single fix.

Effective Go mentions not only exported functions I believe. See: 
```
Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.

// Compile parses a regular expression and returns, if successful,
// a Regexp that can be used to match against text.
func Compile(str string) (*Regexp, error) {
If every doc comment begins with the name of the item it describes, you can use the doc subcommand of the go tool and run the output through grep. Imagine you couldn't remember the name "Compile" but were looking for the parsing function for regular expressions, so you ran the command,
```

So considering this as best practice, we should revisit every comment in all functions and change them to what's is considered best practice.

fredbi:
Hello @abstractj. This PR is actually pushed by a bot. I've seen it on other repos I am working with.
Their tooling actually does a good job in setting comments to get a proper godoc (that is the above mentioned "best practice").
What you are asking for (relevant comments, purged from typos and terrible english in some places) is beyond the capabilities of this bot.

abstractj:
@fredbi thanks for bring this to my attention. I'm closing this PR because we need more than bot recommendations :)

PR #470, opened 2019-03-21
[KEYCLOAK-9870] Fix refresh token issue

fredbi:
Whenever keycloak is set to revoke refresh tokens after usage, we need to get
the new refresh token, its expiry and refresh the cookie in response.


vincentbarat:
We use this patch on our production platform.
Any idea about when it will be integrated?

abstractj:
@vincentbarat Hi Vincent, I'm glad that worked for you. I just returned back from travel and will try to do a proper review/merge it.

abstractj:
@fredbi thank you for the fix, we hope to get it in the next release.

PR #471, opened 2019-03-22
[KEYCLOAK-9880] Make sure the Authorization Header token is only considered when type is Bearer

HansK-p:
Make the function getTokenInBearer only fetch the token when the type is Bearer. This makes it possible to place Keycloak Gatekeeper in front of and fully protect (some) services which use basic authentication. In my case the RabbitMQ web interface.


HansK-p:
Thanks
Not sure if I did this correctly (register issue, etc), but we'll find out. Looks like (almost) exactly the same fix was suggested as a partial solution to issue KEYCLOAK-8856 (closed on Dec 10th 2018 as a duplicate).

nirousseau:
Hello,

We may also handle "malformed" Authorization headers such as :

`
Authorization: Bearer "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"
`

In the example above, the double quotes are breaking the internal Base64 Encoder/Decoder. (`malformed JWS, unable to decode header, illegal base64 data at input byte 0`)

Maybe we can detect them and strip the first and last char of the sequence before decoding ?

nirousseau:
> 
> 
> Hello,
> 
> We may also handle "malformed" Authorization headers such as :
> 
> `Authorization: Bearer "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"`
> 
> In the example above, the double quotes are breaking the internal Base64 Encoder/Decoder. (`malformed JWS, unable to decode header, illegal base64 data at input byte 0`)
> 
> Maybe we can detect them and strip the first and last char of the sequence before decoding ?

Created PR to HansK-p fork in order to implement simple token sanitizer.

fredbi:
@nirousseau  To be honest, I am not a great fan of sanitizing tokens...
 Is there any practical use case where we should allow such malformed headers to pass through? I mean, if we start with such sanitization work, there is no limit to imagination (like stripping trailing '\n' because that guy's test program fails doing so, etc..).

nirousseau:
I'm trying to integrate a young app (from another company) with Gatekeeper. The Javascript of this application is sending Bearer tokens encapsulated with those double quotes while querying the API.

**Root cause :** the issue appears when we the library "auth0/angular-jwt" is used ; while retrieving the Bearer token from local storage (see https://github.com/auth0/angular-jwt/issues/141#issuecomment-405026683).

We can outline that RFC 6750 (https://tools.ietf.org/html/rfc6750#section-2.1) defines the Bearer scheme **without** double quotes.

Anyway, I encountered this issue. The error is not on Gatekeeper side. We don't have to support it, but, it does not cost too much to trim those quotes, even with an additional trim for white chars (space, CR/LF etc.). I will report this issue to the developers of the application I'm deploying.

I won't blame you or get mad if you reject those three lines :-).

@HansK-p contribution is more useful.

nirousseau:
@fredbi My changes are targeting Hansk's feature branch "HansK-p:authorization-header-bearer-token-fetch-fix". Depending on your decision, I will close my own PR to his fork. You will be free to merge this pull request without token sanitization.

bs-matil:
Will it be possible then to keep the old behavior? Otherwise if one has a not protected resource which relies on the gateway headers for authentication. An attacker could just bypass the security mechanism by faking a basic auth header and injecting wrong headers. Or did I got something wrong? If not this should at least not be the default behavior.

HansK-p:
The original patch solves a bug. Currently Keycloak Gatekeeper incorrectly handles a basic auth authorization header token as if it was a JWT token. As a result a https request with a valid JWT token as cookie, but with a Basic Auth header, which is supposed to be handled by the upstream server, is incorrectly rejected by KeyCloak Gatekeeper. 

The current alternatives to patching Keycloak Gatekeeper which I'm aware of are:
1) Rewrite the upstream server (not always an option)
2) Expose the service using basic auth directly on the internet (not always a good option)
3) Use oauth2-proxy or another product (not tested and I prefer Keycloak Gatekeeper)

If the patch makes it possible for an attacker to buypass Keycloak Gatekeeper by faking a basic auth header I assume the same attacker could buypass Keycloak Gatekeeper simply by removing the basic auth header. The only thing this patch does is making sure that Keycloak Gatekeeper ignores an authorization header without a JWT token the same way as it currently ignores a non-existing authorization header. At least as far as I've interpreted the code.

abstractj:
@HansK-p would be nice to add some tests for this, just to be safe. I will follow up on keycloak-dev mailing list to check if others have some thoughts on this.

@bs-matil I believe your concern is addressed in https://tools.ietf.org/html/rfc6750#section-5.1. In summary that's the trade-off about using bearer tokens from my understanding. Unless, I'm missing something.

stianst:
I agree Gatekeeper should only consider header when type is Bearer, and just pass on a basic header. It should not try to sanitize the token though (stripping out invalid ")

HansK-p:
@abstractj It took a few months before I reacted, but the test part is hopefully ok now. I must admit that the test was harder to (re)write than the actual change, so I hope I didn't introduce any mistakes into the test -- but the change to the test has been ok tested :).

abstractj:
@HansK-p thanks for your time on it, will give it a try this week so we can merge.

HansK-p:
@abstractj Thx. I added a couple of more targeted test cases.

PR #473, opened 2019-03-29
[KEYCLOAK-9885] Force cookie lookup for access token and allow raw Authorization headers to be sent "AS IS"

nirousseau:
Hello,

This pull request has been created in order to solve my issue available here : https://issues.jboss.org/browse/KEYCLOAK-9885

The goal if this pull request is to bring more behavior tuning to Keycloak-Gatekeeper :
 - Being able to configure the Keycloak token lookup (for example force cookie lookup for the access token).
 - Being able to forward Authorization header "as-is" to the upstream application (without altering it, transparent way).

**Motivations** are discribed in issue [KEYCLOAK-9885](https://issues.jboss.org/browse/KEYCLOAK-9885).

Tell me if I am required to create the unit tests related to the newly added code.

I'm open to any suggestions / code improvements in order to accept this PR.

Best regards,
Nikita


middleware.go
```diff
@@ -387,7 +387,10 @@ func (r *oauthProxy) identityHeadersMiddleware(custom []string) func(http.Handle
 					req.Header.Set("X-Auth-Token", user.token.Encode())
 				}
 				// add the authorization header if requested
-				if r.config.EnableAuthorizationHeader {
+				if r.config.EnableAuthorizationHeader && r.config.EnableRawAuthorizationHeader {
+					// Forward received Authorization header AS IS
+					req.Header.Set("Authorization", fmt.Sprintf("%s", user.rawAuthorization))
```
golangcibot: S1025: the argument is already a string, there's no need to use fmt.Sprintf (from `gosimple`)

config.go
```diff
@@ -34,35 +34,37 @@ func newDefaultConfig() *Config {
...
-		RequestIDHeader:               "X-Request-ID",
-		ResponseHeaders:               make(map[string]string),
-		SecureCookie:                  true,
-		ServerIdleTimeout:             120 * time.Second,
-		ServerReadTimeout:             10 * time.Second,
-		ServerWriteTimeout:            10 * time.Second,
-		SkipOpenIDProviderTLSVerify:   false,
-		SkipUpstreamTLSVerify:         true,
-		Tags:                          make(map[string]string),
+		AccessTokenDuration:          time.Duration(720) * time.Hour,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

middleware.go
nirousseau: fixed in https://github.com/keycloak/keycloak-gatekeeper/pull/473/commits/1eabb6627a68ea6adcfa7311876b8ea1ad9bcd39

config.go
nirousseau: I'm confused with this review. "time" is imported line 25 of "config.go"

abstractj:
@nirousseau some considerations before people start to review:

1. Before implementing features that you would like to include in Gatekeeper, I'd suggest to first give a heads up on keycloak-dev mailing list with your proposal. That's the way of getting people's feedback.
2. Tests and documentation are mandatory.
3. Please, squash your commits in a single one.

Thanks for submitting this and have a great weekend.

fredbi:
@nirousseau I've had a similar requirement at my place with mixed credentials, e.g. cookies vs token in header.
This normally works fine: if header is present, it is preferred over cookie.

What may be confusing is when token is assumed to be encrypted (e.g. in cookie): then the raw taken in header is no more usable (because gatekeeper tries to decrypt this).

My PR #445 solves this issue, I think in a simple way: config states that access token is _not_ encrypted but we also specify that token _in cookies_ is always encrypted. So this way I could get access either from clients using cookies, or from clients (e.g. apps) using access token in headers.

nirousseau:
Hello @abstractj,

Thank you for the general process for contributing to Gatekeeper.

I have subscribed to to mailing list ; I will send an email asap exposing the challenges I am trying to resolve with this PR.

Documentation is done (code + issue tracker).

I need to write unit test (I'm a newbie with golang).

Thank you for your feedback 💯

nirousseau:
Hello @fredbi,

`This normally works fine: if header is present, it is preferred over cookie.`
Yes, I agree. That's the source of my problem : my upstream application is sending its own token with the Authorization header once the authentication is done (I.e the keycloak access token located in the cookie is verified by the proxy), the proxy will look for the header first (as expected). However, the Authorization header does not contain the access token, so it will drop the token and redirect.

I see two approaches :
- Fallback to the cookie checkup if the authorization header is not valid/present
- Help the proxy by specifying where to look at (my approach with this PR #473 )

We should be able to fallback to cookie inspection if the **Authorization header does not contain the access token from keycloak**. is PR #445 allow this ?

`config states that access token is not encrypted but we also specify that token in cookies is always encrypted. So this way I could get access either from clients using cookies, or from clients (e.g. apps) using access token in headers`
It is quite confusing to specify that tokens are "plaintext" with one parameter, but "cyphered" with another one, but only for cookies. If I'm not mistaken, your PR is more related to encrypted token handling. The changes I submit are not depending on encryption parameters. 

I'm very new to Keycloak, maybe I don't get something. Should I draw a sagittal diagram ?

Anyway, thank you for your feedback.
Regards,
Nikita

fredbi:
> my upstream application is sending its own token 
I think this sentence is very confusing and I can feel some design issue behind this, unless you may clarify this better. Upstream means = behind proxy, so this does not make sense if this app already handles token exchange.

> We should be able to fallback to cookie inspection if the Authorization header does not contain the access > token from keycloak. 
As far as I could test things this is already working like this, only I did not have the case where both cookie and authorization header were present.

> Is PR #445 allow this ?
It allows this specifically for the case when you want access token in cookie to be encrypted, while letting token in header to remain in-the-clear

nirousseau:
> I think this sentence is very confusing and I can feel some design issue behind this, unless you may clarify this better. Upstream means = behind proxy, so this does not make sense if this app already handles token exchange.

Ok I will try to explain it more clearly :thumbsup:

**Background :** The upstream application (the app behind the proxy) does not have official openid support. It relies only on HTTP Headers sent by Gatekeeper. I don't belong to the development team of this application, I'm working with a black box approach.

**End-user authentication with Keycloak** : Once the proxy has validated the request against keycloak (i.e redirection to the SSO login page if no valid session is found), it forwards all requests to the upstream application. Exchanged tokens are stored encrypted in the cookie (kc-access, kc-state). This is the normal behavior, no issues here.

**End-user authentication with the app itself** :
Once the user is landing on the application, forwarded by the proxy, the Javascript of the frontend is querying the application API. It touches a "/login" endpoint that generates a token. Sample : https://gist.github.com/nirousseau/33bb767ada8761fe823fca37d84f94bf
This is a Bearer token that is sent in HTTP Authorization header. It is used by the frontend when querying the API of this application. It contains the user role and business logic information. Since this token is not issued by keycloak, but by the application itself, gatekeeper is confused. This is the root cause of my problem.

> As far as I could test things this is already working like this, only I did not have the case where both cookie and authorization header were present.

I agree. However, from my experiments, when the Authorization Bearer token is set, but not issued by keycloak, gatekeeper does not look for the cookie token. It just fails with `error no session found in request, redirecting for authorization {"error": "missing required claim: sub"}` and redirect the request to the SSO login page. (tested in my environment with enable-encrypted-token=false, Authorization header is set along with tokens in cookie kc-access, kc-state.)

> It allows this specifically for the case when you want access token in cookie to be encrypted, while letting token in header to remain in-the-clear

Ok, I guess this PR will solve partially my issue. As explained above, your PR combined with a mechanism that also looks for the cookie if the Authorization Bearer provided is not valid for gatekeeper (a new case when the exception  `missing required claim: sub` is raised) will address my issue.

Is that clearer ? What strategy do you want I investigate ?

Thank you very much for your support,
Regards,
Nikita

fredbi:
@nirousseau yes much clearer now.

My two cents on this is that the heart of your problem is that several apps are competing for the "Authorization" header. This is a source of confusion indeed, but now I think I understand the point.

I don't think this proxy should be twisted to the point of supporting this kind of mismatches.
Authorization headers passing through this should only be valid, keycloak-issued headers.

IMHO, letting traffic passing through with such headers from other issuers would cause confusion and security issues with other users of this proxy.

There are other ways to work around this particular issue, without twisting the proxy's logics.

* For one, an acceptable solution close to what you did would be to set a config parameter like "cookie-only-auth": then whatever the headers say, cookie is used.
* For another (personally I would have done it this way), upstream passes through a more capable reverse-proxy (e.g. nginx) which rewrites headers back and forth and isolate this specifics of your legacy app: your upstream Authorization header becomes something like X-App-Authorization when received by front-end and sent back to gatekeeper proxy,  then is transformed back to Authorization just before reaching your app

I think the second approach is more flexible, at the cost of an additional proxying, which will eventually disappear when this app has eventually migrated to fully integrated authZ with your IDP.

nirousseau:
Hello,

Indeed, letting the Authorization going through the proxy _can_ be a security issue : we are delegating the authentication role to the application, where the SSO should be the only trusted token issuer.

_Pros/Cons PR473 :_
**Pros :** it is convenient for non-mature applications, we are not required to add another proxy in order to modify headers.
**Cons :** add more complexity to Gatekeeper decision tree, requires more tests, can lead to security issues.

**Mitigation :** We can notify at launch that the option can lead to security issues.

To my mind, I prefer not adding another proxy for my case. Even if the second approach is more flexible, I prefer limiting the number of proxy because of latency issues.

**For now, is this PR still relevant for Gatekeeper as a community contribution ? What can I do ?**

abstractj:
@nirousseau thanks for taking the time working on this, but we do not have any plans to merge this change based on the security implications and all the comments above. 

If you would like to discuss more about, please start a thread on the dev mailing list.

PR #480, opened 2019-06-06
[KEYCLOAK-10668] Do not set the cookie domain

daniel-ac-martin:
By setting the domain attribute on the cookie we were allowing the
cookie to be applied to subdomains where it may not be valid and may
interfere with other services protected by keycloak-gatekeeper.

(For example, a gatekeeper running on https://domain.com could break a
gatekeeper running on https://sub.domain.com .)

Instead, we should simply not set the attribute unless there is
a specific domain configured.

For more information please see section 4.1.2.3 of [RFC 6265].

[RFC 6265]: https://tools.ietf.org/html/rfc6265#section-4.1.2

### DIFF
--- a/cookies.go
+++ b/cookies.go
@@ -28,7 +28,7 @@ import (
 // dropCookie drops a cookie into the response
 func (r *oauthProxy) dropCookie(w http.ResponseWriter, host, name, value string, duration time.Duration) {
 	// step: default to the host header, else the config domain
-	domain := strings.Split(host, ":")[0]
+	domain := ""
 	if r.config.CookieDomain != "" {
 		domain = r.config.CookieDomain
 	}
--- a/cookies_test.go
+++ b/cookies_test.go
@@ -37,7 +37,7 @@ func TestCookieDomainHostHeader(t *testing.T) {
 		}
 	}
 	assert.NotNil(t, cookie)
-	assert.Equal(t, cookie.Domain, "127.0.0.1")
+	assert.Equal(t, cookie.Domain, "")
 }
 
 func TestCookieBasePath(t *testing.T) {
@@ -104,7 +104,7 @@ func TestDropCookie(t *testing.T) {
 	p.dropCookie(resp, req.Host, "test-cookie", "test-value", 0)
 
 	assert.Equal(t, resp.Header().Get("Set-Cookie"),
-		"test-cookie=test-value; Path=/; Domain=127.0.0.1",
+		"test-cookie=test-value; Path=/",
 		"we have not set the cookie, headers: %v", resp.Header())
 
 	req = newFakeHTTPRequest("GET", "/admin")
@@ -113,7 +113,7 @@ func TestDropCookie(t *testing.T) {
 	p.dropCookie(resp, req.Host, "test-cookie", "test-value", 0)
 
 	assert.Equal(t, resp.Header().Get("Set-Cookie"),
-		"test-cookie=test-value; Path=/; Domain=127.0.0.1",
+		"test-cookie=test-value; Path=/",
 		"we have not set the cookie, headers: %v", resp.Header())
 
 	req = newFakeHTTPRequest("GET", "/admin")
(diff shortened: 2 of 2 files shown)


abstractj:
@daniel-ac-martin maybe you missed this, but please read https://github.com/keycloak/keycloak/blob/master/CONTRIBUTING.md and follow these guidelines when contributing to Keycloak

daniel-ac-martin:
@abstractj: Thanks for letting me know. I've raised the issue on the mailing list. Hopefully that will lead to a JIRA ticket being created.

daniel-ac-martin:
Oh, it seems my e-mail to the list was blocked. You guys make it quite hard to contribute!

abstractj:
@daniel-ac-martin not sure what happened with your e-mail, or why it was blocked. Have you subscribed to the mailing list?

About your change, please follow the contribution guidelines and if there's something we can improve to make the process better. Let us now.

daniel-ac-martin:
@abstractj: I wasn't subscribed. - Didn't realise that was a requirement.

Just subscribed and re-sent the e-mail. Hopefully, it will get through this time!

daniel-ac-martin:
@abstractj: JIRA created at https://issues.jboss.org/browse/KEYCLOAK-10668 and this PR has been updated to reference the JIRA.

daniel-ac-martin:
@abstractj: Do you think you will have time to look at this soon?

fredbi:
@daniel-ac-martin as far as I know, setting host as default domain is a sensible default.
I think your feature should rather be explicit, e.g. "DisableCookieDomain".

daniel-ac-martin:
@fredbi: Why would you want the cookie to be applied to subdomains by default? Surely it is safer to limit the cookie to just the hostname?

RobinKnipe:
Howdee! Any chance of this being merged and fixed anytime soon?

fredbi:
@daniel-ac-martin 
> @fredbi: Why would you want the cookie to be applied to subdomains by default? Surely it is safer to limit the cookie to just the hostname?

Yes. I concur. You made a good point. Looks good to me.

abstractj:
@daniel-ac-martin @fredbi one thing concerns me about this change, is this part of RFC:
```
WARNING: Some existing user agents treat an absent Domain
      attribute as if the Domain attribute were present and contained
      the current host name.  For example, if example.com returns a Set-
      Cookie header without a Domain attribute, these user agents will
      erroneously send the cookie to www.example.com as well.
```
Correct me if I'm wrong, but that would open the possibility to malicious attackers to setup a subdomain in the internal network and get the cookies for `example.com` for example.

One alternative to make this change possible is to provide something like:
`--cookie-domain=www.example.com` where users explicitly tell which hosts to the cookie will
   be sent

@fredbi @daniel-ac-martin does it make any sense to you?

daniel-ac-martin:
> Correct me if I'm wrong, but that would open the possibility to malicious attackers to setup a subdomain in the internal network and get the cookies for example.com for example.

Is that not the status quo in anycase?

i.e. This change fixes the problem for modern, RFC-compliant clients but not for non-compliant ones.

Also, if `--cookie-domain=www.example.com` was a thing then a bad actor might still be able to stand up `foo.www.example.com` to steal the cookies.

daniel-ac-martin:
@abstractj: See above. (Forgot to tag you!)

abstractj:
Thanks @daniel-ac-martin, the way how it works today I believe bad actors can do this, at the same time, it is bad for the reasons you mentioned in this PR. But there must be something in between that can make this change possible to keep us safe and at the same time fix the issue you mentioned.

I was thinking about get some inspiration from https://tools.ietf.org/html/rfc4592 about wildcards (we do something similar on Keycloak server for `Web Origins`):

- `--cookie-domain=www.example.com` the cookie will be strictly applied to the domain specified
- `--cookie-domain=*.example.com` we permit cookies to be applied to subdomains

Would that make sense for you?

daniel-ac-martin:
@abstractj: I think the question is how do we get the bad clients to honour any of that?

This change should make things safe for RFC-compliant clients.

For other clients I would think you would need to:
1. Understand what incantation is required to coax out secure behaviour (if any)
2. Employ the incantation based on the presence of a match against a user agent string

Or am I missing something?

abstractj:
@daniel-ac-martin thank you, let's merge it for now. I don't see any issues with this and nobody raised any concerns on the mailing list. So we are good to go.

Thanks for this change.

PR #481, opened 2019-06-13
[KEYCLOAK-10023] Prevent goproxy from deleting middleware headers  

andban:
By default, the goproxy middleware removes all headers from the given ResponseWriter before it adds the headers it has received from the upstream service. This includes all headers injected by the authenticationMiddleware, like a `Set-Cookie` with a refreshed access token.

Setting the `KeepDestinationHeaders` option to `true` should prevent this from happening.

Jira issue: https://issues.jboss.org/browse/KEYCLOAK-10023

### DIFF
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -69,10 +69,11 @@ type fakeRequest struct {
 }
 
 type fakeProxy struct {
-	config  *Config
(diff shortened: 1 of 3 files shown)


andban:
Closing, since #441 already covers this.

PR #487, opened 2019-08-27
Add SSL offloading detection to getRequestHostURL

mmaldonado-wtsh:
When SSL is offloaded by a proxy, ie: Traefik, Nginx, etc., Gatekeeper does not detect that the end-user scheme is https for /oauth/logout (if no redirect value is specified).

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -121,7 +121,7 @@ func getRequestHostURL(r *http.Request) string {
 	}
 
 	scheme := "http"
-	if r.TLS != nil {
+	if r.TLS != nil || r.Header.Get("X-Forwarded-Proto") == "https" {
 		scheme = "https"
 	}
 
--- a/utils.go
+++ b/utils.go
@@ -120,9 +120,9 @@ func getRequestHostURL(r *http.Request) string {
 		hostname = r.Header.Get("X-Forwarded-Host")
 	}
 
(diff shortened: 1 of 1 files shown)


utils.go
```diff
@@ -121,7 +121,7 @@ func getRequestHostURL(r *http.Request) string {
 	}
 
 	scheme := "http"
-	if r.TLS != nil {
+	if r.TLS != nil || r.Header.Get("X-Forwarded-Proto") == "https" {
```
golangcibot: string `https` has 2 occurrences, but such constant `secureScheme` already exists (from `goconst`)

mmaldonado-wtsh: I can switch from `https` to `secureScheme` const, but shouldn't the previous line be fixed from `http` to `unsecureScheme` too?

fredbi: yeah sure it should

abstractj:
@mmaldonado-wtsh I'd appreciate if you take a look at our contribution guidelines. We need tests for this change and also a Jira with the proper steps to reproduce.

abstractj:
I’m closing this issue because it has been inactive for a while. Please reopen if you still find some time to work on this issue. But before implement anything I suggest to discuss enhancements/features on the mailing list.

Thank you!

PR #492, opened 2019-10-31
[KEYCLOAK-10242] Fix setting X-Forwarded-For & X-Real-IP

andrsp:
fixed setting x-forwarded-for headers, add x-real-ip with discovered client’s address.

Fixes KEYCLOAK-10242

Signed-off-by: Andrey Izotikov <andrsp@gmail.com>

### DIFF
--- a/forwarding.go
+++ b/forwarding.go
@@ -40,7 +40,12 @@ func (r *oauthProxy) proxyMiddleware(next http.Handler) http.Handler {
 		}
 
 		// @step: add the proxy forwarding headers
-		req.Header.Add("X-Forwarded-For", realIP(req))
+		req.Header.Set("X-Real-IP", realIP(req))
+		if xff := req.Header.Get(headerXForwardedFor); xff == "" {
+			req.Header.Set("X-Forwarded-For", realIP(req))
+		} else {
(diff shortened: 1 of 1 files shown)


forwarding.go
```diff
@@ -40,7 +40,12 @@ func (r *oauthProxy) proxyMiddleware(next http.Handler) http.Handler {
 		}
 
 		// @step: add the proxy forwarding headers
-		req.Header.Add("X-Forwarded-For", realIP(req))
+		req.Header.Set("X-Real-IP", realIP(req))
+		if xff := req.Header.Get(headerXForwardedFor); xff == "" {
+			req.Header.Set("X-Forwarded-For", realIP(req))
```
aweigold, line 45: This doesn't seem right.  It should append, not replace.

andrsp: It does:
1) If XFF exists in request, forward it to app behind proxy as is.
2) If no XFF received in request, set it equal to client's real IP.

Existing implementation adds new header, so app behind proxy receives two different XFF headers - original one and new with just client's real ip, which is definitely not right too :)

Maybe i missed different cases.. will check.

PR #495, opened 2019-12-10
KEYCLOAK-11499

moremagic:
Add `idp-hint` as a configuration option, so that gatekeeper can set it's own suggested idp.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -85,6 +85,12 @@ func (r *oauthProxy) oauthAuthorizationHandler(w http.ResponseWriter, req *http.
 	}
 
 	authURL := client.AuthCodeURL(req.URL.Query().Get("state"), accessType, "")
+	// idp_hint config
+	idpHint := r.config.IdpHint
+	if idpHint != "" {
+		authURL += "&kc_idp_hint=" + idpHint
+	}
+
 	r.log.Debug("incoming authorization request from client address",
 		zap.String("access_type", accessType),
 		zap.String("auth_url", authURL),
--- a/doc.go
+++ b/doc.go
@@ -153,6 +153,8 @@ type Resource struct {
 
 // Config is the configuration for the proxy
 type Config struct {
+	// IdpHint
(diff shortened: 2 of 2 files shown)


msuret:
It would be useful if `idp-hint` could also be dynamically deduced from the host, e.g. accessing the app from `tenant1.myapp.com` would set `idp_hint` to `tenant1`. What do you think ? I can work on this.

abstractj:
> It would be useful if `idp-hint` could also be dynamically deduced from the host, e.g. accessing the app from `tenant1.myapp.com` would set `idp_hint` to `tenant1`. What do you think ? I can work on this.

Thanks for volunteering to work on this @msuret. Before start working on anything, I'd suggest starting a thread on the dev mailing list, so others can join and provide some feedback on this. 
Also, I didn't have the chance to look at other adapters, but we need to make sure that Gatekeeper has the same behavior as other Keycloak adapters.

handlers.go
```diff
@@ -85,6 +85,12 @@ func (r *oauthProxy) oauthAuthorizationHandler(w http.ResponseWriter, req *http.
 	}
 
 	authURL := client.AuthCodeURL(req.URL.Query().Get("state"), accessType, "")
+	// idp_hint config
+	idpHint := r.config.IdpHint
+	if idpHint != "" {
+		authURL += "&kc_idp_hint=" + idpHint
```
rickyschechter, line 91: Could there sometimes be other url parameters already in authURL? I am thinking "&" vs "?"
If that is never a possibility please disregard, otherwise I suggest adding some pre-inspection of authURL before appending.

rickyschechter:
Thank you @abstractj and @moremagic ! The change looks simple enough. 

Line (91) I added a comment. I agree that tests are missing still.

Could you please point me at where I can download the built artifact? I could probably test it manually and report based on my use/findings.

abstractj:
@moremagic would you like to follow up on @rickyschechter's comments? Do you have any plans to introduce tests?

PR #496, opened 2019-12-10
[KEYCLOAK-12427] Improved testing using the current time

moremagic:

### DIFF
--- a/utils_test.go
+++ b/utils_test.go
@@ -438,22 +438,25 @@ func TestFileExists(t *testing.T) {
 }
 
 func TestGetWithin(t *testing.T) {
+	Now = func() time.Time { return time.Date(2019, time.December, 25, 0, 0, 0, 0, time.UTC) }
 	cs := []struct {
 		Expires  time.Time
 		Percent  float64
 		Expected time.Duration
 	}{
 		{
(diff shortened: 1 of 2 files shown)


abstractj:
@moremagic thanks for submitting this, as mentioned in the same Jira I could not reproduce the steps described there. Could you please provide more details?

utils_test.go
```diff
@@ -438,22 +438,25 @@ func TestFileExists(t *testing.T) {
 }
 
 func TestGetWithin(t *testing.T) {
+	Now = func() time.Time { return time.Date(2019, time.December, 25, 0, 0, 0, 0, time.UTC) }
 	cs := []struct {
```
abstractj: I believe that assign this function to `Now` would be misleading, because may give people the impression that this refers to the current time, while it's not.

@slaskawi what do you think about these changes?

abstractj:
@moremagic thanks for your time work on this, but we could not reproduce the issue reported on https://issues.redhat.com/browse/KEYCLOAK-12427. Please take a look at the comments there and let us know what you think on the dev or user mailing list.

For now, I'm closing this. But if still, you would like to move forward with this change, do not hesitate to start a discussion on the mailing list.

PR #500, opened 2020-01-23
Verify upstream certificate by default

daniel-ac-martin:
Hi guys,

Just spotted this and wanted to _quickly_ raise it. (You've disabled issues so it's not easy to do so!)

Is this really a good default to have? Wouldn't it be better to verify by default and only skip verification if the user explicitly demands it?

Apologies if I'm missing something!

### DIFF
--- a/config.go
+++ b/config.go
@@ -62,7 +62,7 @@ func newDefaultConfig() *Config {
 		ServerReadTimeout:             10 * time.Second,
 		ServerWriteTimeout:            10 * time.Second,
 		SkipOpenIDProviderTLSVerify:   false,
-		SkipUpstreamTLSVerify:         true,
+		SkipUpstreamTLSVerify:         false,
 		Tags:                          make(map[string]string),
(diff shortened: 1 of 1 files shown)


abstractj:
Hi @daniel-ac-martin, GH issues are disabled because we need to keep repositories consistent across the Keycloak organization. What you're suggesting makes sense to me, but I think worth to start a thread on the dev mailing list (https://groups.google.com/forum/#!forum/keycloak-dev) so we can get feedback from others.

abstractj:
@daniel-ac-martin would you like to move forward with these changes? If yes, please take a look at our contributing guidelines, we need a Jira and also would be nice to start a thread on the dev mailing list.

abstractj:
@daniel-ac-martin thanks for submitting this. For now I'm closing this due to the lack of activity, if you would like to file a Jira or start a discussion about this on the dev mailing list, please reopen.

PR #501, opened 2020-01-31
[WIP] forward auth mode

kolsys:
Please check.

I've added “forward auth” mode and support for filtering by proxying hostname.
In order to activate forward auth mode you need set upstream-url: **none**
In this mode gatekeeper will not proxy request and answer 200 or 403 error code. It’s useful in balancers that support forward auth.  [Traefik](https://github.com/containous/traefik) for example.
Also I've added support for filtering by hostname. In this mode you can specify request hostname in URI.
```
hostname-mode: true
resources:
- uri: /my.app.com/admin*
  roles:
  - client:admin
- uri: /my.app.com/*
  roles:
  - user
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -189,8 +189,14 @@ func (r *oauthProxy) createReverseProxy() error {
 		})
 		engine.Use(c.Handler)
 	}
-
-	engine.Use(r.proxyMiddleware)
+	// proxying request only if not forward auth mode
+	if r.config.Upstream != forwardAuthUpstream {
+		engine.Use(r.proxyMiddleware)
+	}
+	// modify request in hostname mode
+	if r.config.HostnameMode {
+		engine.Use(r.hostnameMiddleware)
+	}
 	r.router = engine
 
 	if len(r.config.ResponseHeaders) > 0 {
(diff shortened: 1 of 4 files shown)


server.go
```diff
@@ -190,7 +190,12 @@ func (r *oauthProxy) createReverseProxy() error {
 		engine.Use(c.Handler)
 	}
 
-	engine.Use(r.proxyMiddleware)
+	if r.config.Upstream != forwardAuthUpstream {
+		engine.Use(r.proxyMiddleware)
+	}
+	if r.config.HostnameMode {
```
golangcibot: if statements should only be cuddled with assignments (from `wsl`)

GerkinDev:
I'd be happy to see this. It would greatly simplify the use of authentication proxy with traefik (see the docs of their [forward auth middleware](https://docs.traefik.io/middlewares/forwardauth/)

abstractj:
@GerkinDev please file a new issue with all the details and use case scenario. As you may noticed the author closed the PR, so we need a new discussion about this.

kolsys:
@GerkinDev it's realised in #505 
We are using it in "Hostname" mode and refused to use traefik. But the forward auth mode is implemented.

PR #508, opened 2020-02-28
[KEYCLOAK-13165] Add Host into header

xcffl:
Add Host to the forwarding headers to make a more standard behavior

### DIFF
--- a/forwarding.go
+++ b/forwarding.go
@@ -40,6 +40,7 @@ func (r *oauthProxy) proxyMiddleware(next http.Handler) http.Handler {
 		}
 
 		// @step: add the proxy forwarding headers
+		req.Header.Set("Host", req.Host)
 		req.Header.Add("X-Forwarded-For", realIP(req))
 		req.Header.Set("X-Forwarded-Host", req.Host)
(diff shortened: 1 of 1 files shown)


xcffl:
It's tt-rss which evaluate the received `Host` header. I've notice some other services do so.

As said in https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypreservehost, 
> This option should normally be turned `Off`. It is mostly useful in special configurations like proxied mass name-based virtual hosting, **where the original Host header needs to be evaluated by the backend server.**

Actually it is the case. I believe a lot of backends does not have a reverse-proxy-aware mode whereas they check the `Host` header for security reasons.

stianst:
An option to preserve the host header makes perfect sense. However, in this PR you have it hard-coded which is not correct. It needs to be a feature that is disabled by default. If you would like to contribute this feature please:

* Create a JIRA feature request with explaining the feature and motivation of the feature
* Send a PR with the feature that is disabled by default, including tests
* Send a PR to the documentation to document the feature

I'm going to close this PR for the above reasons, but if you would like to contribute this as an optional feature that would be great.

PR #511, opened 2020-03-26
[KEYCLOAK-12100] Custom Logging Time Format

Madankapoor:
Initial design idea to add Custom Logging Time Format. 
As per the existing design, we are setting up a production configuration by default. 
 we have below mentioned flags related to zap logging configuration.
- enable-json-logging:  To switch to JSON or back to console(text) logging.
- disable-all-logging: Disable All Logging
- verbose: Uses Development and Debug Logging Level. 

I propose we add below extra configuration parameters. 
- logging-time-format: logging time format can be set to epoch, millis, nano, or iso8601. (Added in Pull Request )

### DIFF
--- a/server.go
+++ b/server.go
@@ -45,7 +45,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/cors"
 	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 )
 
 type oauthProxy struct {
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -148,7 +148,8 @@ func createLogger(config *Config) (*zap.Logger, error) {
 		c.Development = true
 		c.Level = zap.NewAtomicLevelAt(zap.DebugLevel)
 	}
-
+	// Set Time Logging Format
+	c.EncoderConfig.EncodeTime.UnmarshalText([]byte(config.LoggingTimeFormat))
```
golangcibot: Error return value of `c.EncoderConfig.EncodeTime.UnmarshalText` is not checked (from `errcheck`)

Madankapoor: Fixed it

PR #533, opened 2020-04-17
Enable GZIP compression for keycloak-gatekeeper

13rentgen:
* Implement GZIP compression for all incoming requests

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -16,11 +16,15 @@ limitations under the License.
 package main
 
 import (
+	"compress/gzip"
 	"context"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	uuid "github.com/gofrs/uuid"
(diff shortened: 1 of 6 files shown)


middleware.go
```diff
@@ -18,6 +18,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/NYTimes/gziphandler"
```
golangcibot, line 21: File is not `goimports`-ed (from `goimports`)
```suggestion

```

abstractj:
@stianst this PR introduces another extra dependency only to achieve gzip compression and I'm really not sure if this is something we would like to introduce on Louketo. Maybe, might work to introduce it as an option, instead of making it the default. For example: `--enable-compression`.

Anyways, if we agree to move forward with these changes, we need tests for it.

stianst:
I'd say GZIP compression is something we should support right, so picking the correct library for the task seems like the way to go.

abstractj:
> I'd say GZIP compression is something we should support right, so picking the correct library for the task seems like the way to go.

+1

13rentgen:
> Could you please refactor to https://golang.org/pkg/compress/gzip/ ?

Refactored. Disable compression by default, will be enabled on 'enable-compression'. Added tests.

abstractj:
Thank you @13rentgen!

PR #535, opened 2020-04-18
KEYCLOAK-10864 Pass URI path upstream verbatim

### LINKED ISSUE #528: Gatekeeper mangles/truncates proxied request, even when encoded
Gatekeeper 9.0.2 is mangling requests it proxies in certain circumstances. Specifically, when put in front of [Jenkins](https://github.com/jenkinsci/jenkins), it causes Jenkins' [reverse proxy validator](https://wiki.jenkins.io/display/JENKINS/Jenkins+says+my+reverse+proxy+setup+is+broken) to fail. 

I do not know why Jenkins performs the test the way it does, but it has been around for years and is not going to change. The browser (via AJAX) hits an endpoint. The endpoint redirects the request to another endpoint URL that includes the URL-encoded value of the "Referrer" header it received. The second endpoint validates that the Referrer matches the expected value and returns an error if it does not.

Thus Gatekeeper receives a request header like:
```
GET /administrativeMonitor/hudson.diagnosis.ReverseProxySetupMonitor/testForReverseProxySetup/https%3A%2F%2Fjenkins.example.com%2Fmanage/
```
Note the `%3A%2F%2F` which decdes to `://` (colon slash slash).
However, Gatekeeper decodes the request and squishes the double slash into one in the [log](https://github.com/keycloak/keycloak-gatekeeper/blob/d7f64e71bea70dc4023dbd927983962c4e255ac5/middleware.go#L90):
```
keycloak-gatekeeper/middleware.go:90  client request {..."method": "GET", "
path": "/administrativeMonitor/hudson.diagnosis.ReverseProxySetupMonitor/testForReverseProxySetup/https:/jenkins.example.com/manage/"}
```
and by the time Jenkins receives it, Jenkins parses the request as 
```
/administrativeMonitor/hudson.diagnosis.ReverseProxySetupMonitor/testForReverseProxySetup/https:
```
Everything after the colon has been dropped. 

Gatekeeper should not be altering the request. Any processing it needs to do on the request to parse it, validate it, apply security rules, etc., should be done on a copy of the request string. The original request string should be forwarded if the request is allowed. 

I am guessing that https://github.com/keycloak/keycloak-gatekeeper/blob/d7f64e71bea70dc4023dbd927983962c4e255ac5/middleware.go#L51 is likely the source of the problem. 

I do not understand the point of [normalizing the URL](https://github.com/keycloak/keycloak-gatekeeper/blob/d7f64e71bea70dc4023dbd927983962c4e255ac5/middleware.go#L44) in the request to begin with, since it can lead to problems like what we see here with Jenkins, but at the very least I think this line replacing the `RawPath` with the normalized `Path` should be removed. Likewise, the `RawPath` and `RequestURI` should not be replaced with the parsed `Path` at the end of the function here https://github.com/keycloak/keycloak-gatekeeper/blob/d7f64e71bea70dc4023dbd927983962c4e255ac5/middleware.go#L65-L66

Nuru:
- Fixes [KEYCLOAK-10864](https://issues.redhat.com/browse/KEYCLOAK-10864) which will also fix [KEYCLOAK-11276](https://issues.redhat.com/browse/KEYCLOAK-11276) and [KEYCLOAK-13315](https://issues.redhat.com/browse/KEYCLOAK-13315)
- Closes #528
- Closes #483 by superseding it as a bug fix, not an option

No documentation needed, as this is a bug fix. 

### Testing
Very importantly, it passes all the existing unit tests, plus new ones I wrote with URLs associated with open bugs. I also corrected the test rig to properly test the URLs that are being passed upstream rather than the URLs sent to Gatekeeper for proper values.

Of course, the question is, does it also fix the Jenkins problem, and that it does. 

If you use the test setup at https://github.com/primeroz/keycloak-10864-test-jenkins (not written by me) you will see the error message:

It appears your reverse proxy setup is broken

Then, in the docker-compose.yml, replace the proxy image "quay.io/keycloak/keycloak-gatekeeper:8.0.1" with "nuru/key-gate:KEYCLOAK-10864" (which contains the `keycloak-gatekeeper` binary from this PR) and then docker down, docker up, and you will see the error message is gone. 

See discussion on [Keycloak Dev](https://groups.google.com/forum/#!topic/keycloak-dev/GMFu1Q61PMg) but please add comments here.

### DIFF
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -334,16 +334,17 @@ func TestMetricsMiddleware(t *testing.T) {
 	cfg.LocalhostMetrics = true
 	requests := []fakeRequest{
 		{
-			URI:                     cfg.WithOAuthURI(metricsURL),
-			ExpectedCode:            http.StatusOK,
-			ExpectedContentContains: "proxy_request_status_total",
-		},
-		{
 			URI: cfg.WithOAuthURI(metricsURL),
 			Headers: map[string]string{
 				"X-Forwarded-For": "10.0.0.1",
 			},
 			ExpectedCode: http.StatusForbidden,
+		},
+		// Some request must run before this one to generate request status numbers
+		{
+			URI:                     cfg.WithOAuthURI(metricsURL),
+			ExpectedCode:            http.StatusOK,
+			ExpectedContentContains: "proxy_request_status_total",
 		},
 	}
 	newFakeProxy(cfg).RunTests(t, requests)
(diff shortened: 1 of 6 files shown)


middleware_test.go
```diff
@@ -433,6 +434,100 @@ func TestMethodExclusions(t *testing.T) {
 	newFakeProxy(cfg).RunTests(t, requests)
 }
 
+func TestPreserveURLEncoding(t *testing.T) {
```
golangcibot, line 437: Function 'TestPreserveURLEncoding' is too long (91 > 60) (from `funlen`)

server.go
```diff
@@ -21,6 +21,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"go.uber.org/zap/zapcore"
```
golangcibot, line 24: File is not `goimports`-ed (from `goimports`)
```suggestion

```

middleware_test.go
Nuru: 8 other functions in `middleware_test.go` are also > 60 lines. It's because the test cases are long. I suggest living with it and keeping the tests together rather than splitting them up.

server.go
Nuru: If you are going to check `goimports`, I suggest adding it to the Makefile either as part of `format` or as a separate target.

Nuru:
@abstractj Please consider this PR as an improvement over #483 (details on the [dev list](https://groups.google.com/forum/#!topic/keycloak-dev/GMFu1Q61PMg))

stianst:
If this PR is using the original path when invoking upstream, while using the normalized path when applying permission checks, then I'm in favour of this approach.

Nuru:
> If this PR is using the original path when invoking upstream, while using the normalized path when applying permission checks, then I'm in favour of this approach.

@stianst That is it exactly.

stianst:
+1 On this approach from me - haven't looked at the code though

abstractj:
@Nuru thank you so much for submitting this and apologies if it took longer than expected. Your PR was superseded by https://github.com/louketo/louketo-proxy/pull/626, which is pretty much a rebase of your changes.

If you have any questions, please let me know.

Issue #516, opened 2020-04-03
Migrate the codebase from Gatekeeper repository


abstractj:
## What

Move the codebase from https://github.com/keycloak/keycloak-gatekeeper to https://github.com/louketo

## Why

We agreed that would be a good idea to use Gatekeeper codebase as a starting point for development, so we could iterate over it.

## How

Move the codebase to louketo repository by preserving Git history

stianst:
I thought we where going to just move the repo? Not clone it.

abstractj:
@stianst fixed :)

jangaraj:
Are there any public details about future of this repo? Why it will be moved to Louketo and what is Louketo?

mreinli:
Yes, would also be interesting to hear about the plans here.

abstractj:
@jangaraj @mreinli please give me few hours, I will follow up on the dev mailing list.

abstractj:
@jangaraj @mreinli please let me know if it helps to clarify this issue https://groups.google.com/forum/#!topic/keycloak-dev/oDyw94BWxM0. If not, please do not hesitate to ask me any questions on the same thread.

abstractj:
Resolved by https://github.com/louketo/louketo-proxy/pull/603

PR #603, opened 2020-05-04
Migrate the codebase from Gatekeeper repository #516

### LINKED ISSUE #541: `make docker-build` and `make docker` do not work on 9.0.2 branch
`make docker-build` does not work, because it is hardcoded to Go 1.10, which doesn't seem to have support for pulling dependencies. Updating `GOVERSION` to 1.14.2 in `Makefile` does the trick.

`make docker` does not work because there is no `Dockerfile`.

A minimal working `Dockerfile` might look like this (if you add the `static` target as a dependency for `docker` in `Makefile`)

```
FROM alpine:3.11
ADD bin/keycloak-gatekeeper /usr/bin/keycloak-gatekeeper
ENTRYPOINT /usr/bin/keycloak-gatekeeper
```

abstractj:
In the past month, [we discussed](https://groups.google.com/forum/#!msg/keycloak-dev/oDyw94BWxM0/zc0J9R10BwAJ) the idea of having a new home for Gatekeeper, based on that we decided to create a new
organization on GitHub and rename the project to [Louketo Proxy](https://www.wordhippo.com/what-is/the-meaning-of/greek-word-b1b9dc52a1b1d327a31a9684bd7c34f9af80baf1.html).

This pull request refactors the codebase to the new name and should be merged only after the transfer
of the repository.

## Some considerations

* The documentation will be improved, but it's out of the scope for this pull-request
* Migration of the [Dockerfile](https://github.com/keycloak/keycloak-containers/tree/master/gatekeeper) after the transition, that also covers #541.
* There's a lot of work to do this week, so please be patient :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 
 *.swp
 .idea/
+.vscode/
 bin/
 vendor/
 release/
@@ -11,6 +12,7 @@ test.sock
 tests/redis.conf
 tests/*.csr
 *.orig
+*.code-workspace
 debug
 
 *.iml
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,8 @@ language: go
 
 matrix:
   include:
+  - go: 1.13.x
   - go: 1.12.x
-    env: GO111MODULE=on
-  - go: 1.11.x
     env: GO111MODULE=on
     
 install:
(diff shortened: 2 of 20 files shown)


abstractj:
@stianst thanks for reviewing. Ready for another round.

abstractj:
Resolves #516

Issue #530, opened 2020-04-15
 You have not specified the listening interface - error


rogerioefonseca:
I know that I only have to pass the config parameter, but would be nice to document it on the README like bitnami made. 

https://github.com/bitnami/bitnami-docker-keycloak-gatekeeper#using-a-configuration-file

bpasson:
Feel free to assign it to me. I already submitted a PR for it as you can see.

abstractj:
Resolved by #608

PR #608, opened 2020-05-07
#530: Updated docker run command in README.md

bpasson:
A PR for solving #530


abstractj:
@ASzc whenever you have the chance this week, could you please check if I didn't mess things up here https://quay.io/repository/louketo/louketo-proxy?tab=tags ? From what I checked, things are working as expected.

ASzc:
@abstractj Seems ok on Quay, although we may need further updates to the Makefile. `make container-test` for example pulls `docker.io/louketo/louketo-proxy:v2.3.0`, which doesn't exist.

Issue #615, opened 2020-05-14
Automate Louketo releases


abstractj:
Louketo releases should be automated make this task easier and less error prone.

## Why?

We should not do any manual steps to publish a new release, other than just press a button.

## How

* By combining GitHub Actions with Goreleaser we can easily achieve that

## Acceptance criteria

* After draft a new release and publish it, all the artifacts for macOS, Linux and Windows should be 
available in the release section after the build
* Releases should be blocked if tests are broken 

## Additional Information

* [GoReleaser](https://goreleaser.com/)

PR #616, opened 2020-05-15
Automate Louketo releases (#615)

abstractj:
# Testing this PR

* Download the `*.yml` files from this PR and add it to `.github/workflows` into the fork of
Louketo
* Push the changes to `master` at your fork
* Go to https://github.com/<yourhandle>/louketo-proxy/releases
* Draft a new release
* Specify the tag version
* Publish the release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,33 @@
+name: Release
+
+# Only trigger this action when a new release is published
+on:
+  release:
+    types: [published]
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      # Setup the workflow to use the specific version of Go
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.13.1'
+      # Checkout the repository
+      - name: Checkout
+        uses: actions/checkout@v2
+      # Cleans up unreachable Git objects and convert a shallow clone to a regular one if needed
+      - name: Unshallow
+        run: git fetch --prune --unshallow
+      # Run tests prior to release    
+      - name: Test
+        run: make test
+      # Publish the binaries
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,57 @@
+# Mark the release as not ready for production in case there is an 
+# indicator for this in the tag e.g. v1.0.0-rc1
+# See: https://goreleaser.com/customization/release/
+release:
+  prerelease: auto
+# Prune any dependencies that are no longer needed before the build starts
+before:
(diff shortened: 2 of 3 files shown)


stianst:
Nice. LGTM as a starting point. In time you may need to add more things to a release like publish containers, update website, etc..

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/checkout@v2
+      # Cleans up unreachable Git objects and convert a shallow clone to a regular one if needed
+      - name: Unshallow
+        run: git fetch --prune --unshallow
+      # Run tests prior to release    
+      - name: Test
+        run: make test
+      # Publish the binaries
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
```
stianst, line 28: Does this thing also end up creating tags and uploading bits to github releases?

abstractj: @stianst that tag is created by us when, after drafting a new release, we publish it. GoReleaser will be responsible to buildm publish the binaries and generate the changelog.

abstractj:
Converted to draft because I will add documentation to the process and also record a demo.

abstractj:
@stianst added a releases doc and also a video with a demo.

abstractj:
@ASzc I totally understand your point, but we still have the `Makefile` which is the "command-line" way of doing this. The purpose of this PR is to reduce the complexity while doing releases, but we can improve the process and discuss the productization details later on.

Meanwhile, I'd appreciate your review and if you could give it a try and report if something is not working on your end.

ASzc:
@abstractj GitHub doesn't seem to like it if one pushes in the workflows from outside. I tried to add in a workflow through the online editor too, and it doesn't seem to actually add the workflow. I only have the initial example workflow.

Drafting a release doesn't seem to attach binaries automatically: https://github.com/ASzc/louketo-proxy/releases/tag/test

abstractj:
> @abstractj GitHub doesn't seem to like it if one pushes in the workflows from outside. I tried to add in a workflow through the online editor too, and it doesn't seem to actually add the workflow. I only have the initial example workflow.
> 
> Drafting a release doesn't seem to attach binaries automatically: https://github.com/ASzc/louketo-proxy/releases/tag/test

@ASzc I believe that happened because of the specified version, which is a non valid Semver. See the error [from your logs](https://github.com/ASzc/louketo-proxy/runs/734843669?check_suite_focus=true):

```
 release failed after 0.34s error=failed to parse tag test as semver: Invalid Semantic Version
```
Could you please give it another try with something like `X.Y.Z` and if that does not work we can jump in a call to check what's wrong.

.github/workflows/release.yml
abstractj: @stianst The only manual steps you need to do is to `Draft a new release` and once you do that, you have a new tag and need to wait for the binaries to be published.

Tagging after the release is something we could improve and explore, but I don't see any other way of doing this, if we want to just press a button. Another alternative, would be to do that through the command line (more error prone).

ASzc:
@abstractj That seems to have worked [here](https://github.com/ASzc/louketo-proxy/releases/tag/1.33.7)

PR #625, opened 2020-05-27
Added fallback if no realm roles were detected; It now extracts the r…

FireDrunk:
First of all, some explanation is required for this PR I think. I'm trying to get the proxy working with Azure App Registrations (or Enterprise Applications, whatever you want to call them).

Azure supports something they call AppRoles. How you manage them is not very relevant, but suffice to say, it results in a "roles" claim in the ID token. To support this usecase in louketo, I've added a fallback to the user_context that tries to extract the roles claims *after* it fails to load the realm roles. This is the easiest fixt.

If you find this acceptable, that would of course be nice, but maybe it's better to have a commandline option to control this behaviour. I don't think it's bad behaviour, and it won't break anything since the roles claim cannot be forged and is within trusted context. If no realm roles were found, it means Keycloak is not in the picture, and we are just using plain OIDC.

# Title
Roles claim fallback

## Summary 
This adds an extra parse step of the claims in the ID token to 'find' roles.

## Type

[] Bug fix
[] Feature request
[x] Enhancement
[] Docs

## Why?
It 'fixes' roles functionality coming from Azure AD AppRoles.

## Requirements
A working App Registration / Enterprise Application with AppRoles setup.
## How to try it?
Guide: https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps
## Documentation
I don't think it's required just yet. But if so, I can write some.

## Additional Information

It's my first PR for this repo, please be gentle :)

## Checklist:
- [ ] My change requires a change to the documentation or CHANGELOG.
- [ ] I have updated the documentation/CHANGELOG accordingly.

### DIFF
--- a/user_context.go
+++ b/user_context.go
@@ -63,6 +63,15 @@ func extractIdentity(token jose.JWT) (*userContext, error) {
 		}
 	}
 
+	// @step: fallback to pure roles extraction if no realm roles were detected
+	if len(roleList) == 0 {
+		if roles, found := claims[claimResourceRoles]; found {
+			for _, r := range roles.([]interface{}) {
+				roleList = append(roleList, fmt.Sprintf("%s", r))
+			}
+		}
+	}
+
 	// @step: extract the client roles from the access token
 	if accesses, found := claims[claimResourceAccess].(map[string]interface{}); found {
 		for name, list := range accesses {


abstractj:
@FireDrunk thanks so much for spending time on this. However, I'm not sure if this is something we would like to add right now because it introduces specific checks for Azure into the codebase. So any specifics other provides should be extracted to a separate file.

I still want to try the steps you mentioned and would like to understand more about your motivation. Honestly, I'm slightly confused, you mentioned that this is _the easiest fix._, but it is marked as an "enhancement". Is this a bug you found while trying Azure with Louketo?

What would be the worst thing that could happen if we don't introduce this patch? 

Another thing that worth to mention, is the fact that we need tests and documentation. But this can wait until we discuss the this PR

FireDrunk:
Good question, the problem is that the regular code (the first code block above my changes) is Keycloak specific. It checks for the realm specific claims here:
https://github.com/louketo/louketo-proxy/blob/master/user_context.go#L58

This means, that it's bound to Keycloak. Your premise is that this should be seperated into seperate files, which in itself a good idea. This would mean there are specific 'types' of tokens that Louketo recognises. We could make a parser that tries different sets of parsing rules, and add commandline options to set your preference. BUT, this would mean (in my opinion) that we should also move the Keycloak specific parsing to a seperate file.

This is where my explanation of 'easiest' fix came from. The code now let's the realm roles by parse the keycloak way, and if (and only if) that fails (the roles are ampty after that check), it tries the direct approach. This is not entirely Azure specific. It just looks for a 'roles' claim that is not nested.

If you want, I'm happy to (try to) build a basic setup for a multiple token types and a parser.
Problem is, that that is a much bigger (and design heavy) change, and I was kind of hoping this would be merged, because we want to use this in our testing environment ;)

FireDrunk:
Maybe implementing something like this: https://www.tutorialspoint.com/design_pattern/filter_pattern.htm (but in Go offcourse...)

FireDrunk:
@abstractj Do you have any time to look at my proposal?

abstractj:
> Good question, the problem is that the regular code (the first code block above my changes) is Keycloak specific. It checks for the realm specific claims here:
> https://github.com/louketo/louketo-proxy/blob/master/user_context.go#L58
 
@FireDrunk thanks for looking into this. The specifics for Keycloak should be refactored as we've been working to make Louketo work with more providers. There's still a lot of work to do, but that's our ultimate goal and the API design is still something to be discussed.

> 
> This means, that it's bound to Keycloak. Your premise is that this should be seperated into seperate files, which in itself a good idea. This would mean there are specific 'types' of tokens that Louketo recognises. We could make a parser that tries different sets of parsing rules, and add commandline options to set your preference. BUT, this would mean (in my opinion) that we should also move the Keycloak specific parsing to a seperate file.
> 
> This is where my explanation of 'easiest' fix came from. The code now tries to find the realm roles by parsing the token the keycloak way, and if (_and only if_) that fails (the roles are ampty after that check), it tries the direct approach. This is not entirely Azure specific. It just looks for a 'roles' claim that is not nested. This means that people adding 'custom' roles can just use a basic claim called 'roles'. There are several OIDC providers that support adding a custom claim called 'roles'
> 
> If you want, I'm happy to (try to) build a basic setup for a multiple token types and a parser.
> Problem is, that that is a much bigger (and design heavy) change, and I was kind of hoping this would be merged, because we want to use this in our testing environment ;)

I believe you are correct in most of the ideas and indeed, it requires proper design and more work. I understand that it can be frustrating because you need this right now. At the same time, if we start to add more and more code without a proper design discussion, the codebase can become bloated and hard to maintain.

That being said +1 for having a proper design discussion on how to properly support more providers, but I don't see us moving forward with this PR as is.@FireDrunk thanks for looking into this. The specifics for Keycloak should be refactored as we've been working to make Louketo work with more providers. There's still a lot of work to do, but that's our ultimate goal and the API design is still something to be discussed.

@stianst what do you think?

stianst:
Supporting multiple providers as well as different token formats is something we need to support. What could be good is to have the "core" require configuration to parse tokens rather than have it hard-coded, so it's be something like "--role-claim=claim-name". Then we can have vendor profiles where these things are pre-configured, something like "--vendor=kc".

FireDrunk:
Ok, that seems like a good idea. I'm willing to invest some time in this to make it work.
Is there any way we can 'agree' on a design, so I can start working on this? It would be a waste to spend a lot of time, and have the PR be denied or require rework multiple times.

I think my previous suggestion on having specific files for specific parsing rules  for claims fits @stianst suggestion. We can have something like:
./vendor.go: Containing vendor includes (it includes all structs, but only executes those requested via the cmd line.
./vendor/<vendorname>/roles.go: Parse the roles claim into the user context
./vendor/<vendorname>/groups.go: Parse the groups claim into the user context

We can write tests for each individual vendor in a vendor_test directory.

stianst:
Not sure how @abstractj wants to deal with design discussions, but I would suggest at least we try GitHub discussions, or just use an issue for it. At least we should have a complete design somewhere that we can agree on, and that will as you say prevent having to redo PRs.

FireDrunk:
Ok, do you want me to start a discussion in a new Issue, and close this PR?

abstractj:
@stianst +1 let use GH issues for design discussions and centralize everything there. @FireDrunk sure let's close this PR and work on the design, in this way we make a better usage of your time. Thanks for understanding.

PR #626, opened 2020-05-29
Pass URI path upstream verbatim

### LINKED ISSUE #528: Gatekeeper mangles/truncates proxied request, even when encoded

abstractj:
Closes #528
Closes #483 by superseding it as a bug fix, not an option

This PR is just a rebase of https://github.com/louketo/louketo-proxy/pull/535

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -385,6 +385,11 @@ type RequestScope struct {
 	AccessDenied bool
 	// Identity is the user Identity of the request
 	Identity *userContext
+	// The parsed (unescaped) value of the request path
+	Path string
+	// Preserve the original request path: KEYCLOAK-10864, KEYCLOAK-11276, KEYCLOAK-13315
+	// The exact path received in the request, if different than Path
+	RawPath string
 }
 
 // storage is used to hold the offline refresh token, assuming you don't want to use
--- a/forwarding.go
+++ b/forwarding.go
@@ -32,8 +32,9 @@ func (r *oauthProxy) proxyMiddleware(next http.Handler) http.Handler {
 
 		// @step: retrieve the request scope
 		scope := req.Context().Value(contextScopeName)
+		var sc *RequestScope
 		if scope != nil {
-			sc := scope.(*RequestScope)
+			sc = scope.(*RequestScope)
 			if sc.AccessDenied {
 				return
 			}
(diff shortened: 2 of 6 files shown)


abstractj:
@stianst merging this as we already agreed on https://github.com/louketo/louketo-proxy/pull/535

PR #628, opened 2020-06-02
Use HTTP 303 for redirects, instead HTTP 307

### LINKED ISSUE #627: Why does louketo send User Agents to the authorization server with 307?
## Summary 

Was testing a protected endpoint using POST, and I was redirected to `/oauth/authorize` on louketo with a 307, and then subsequently redirected to my authorization server (Keycloak) with a 307.

## Environment

Looking at https://github.com/louketo/louketo-proxy/blob/f623dee81736ee53ab3ec531b550496a76b8231f/handlers.go#L70

(Not environment specific)

## Expected Results

I thought we should be sending User-Agents to the authorization server with a 303, as 303 hints the User-Agent to use a GET (https://tools.ietf.org/html/rfc2616#section-10.3.4)

I think the method itself isn't so much of an issue, but sending a 307 suggests that the User-Agent [can repeat the original request on the new URI](https://tools.ietf.org/html/rfc2616#section-10.3.8), which might leak something to the authorization provider.

## Actual Results

Louketo tries to send the User-Agent to the authorization server with a 307.

## Steps to reproduce

Set up louketo to proxy a protected endpoint and try POSTing to an endpoint requiring authentication

## Additional Information

Related question: In [redirectToAuthorization](https://github.com/louketo/louketo-proxy/blob/f623dee81736ee53ab3ec531b550496a76b8231f/misc.go#L94), why is the default 307? I was thinking that there isn't really a reason to have 307, since 307 would encourage the User-Agent to repeat the original request on the new URI, with the original method and data, etc, which probably isn't necessary since louketo's /oauth/authorize endpoint isn't going to make use of all that.

abstractj:
The option `InvalidAuthRedirectsWith303` was removed, as does not make
sense to have it anymore.

More details:
https://tools.ietf.org/id/draft-ietf-oauth-security-topics-08.html#rfc.section.3.9

Resolves #627

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -306,8 +306,6 @@ type Config struct {
 	// EncryptionKey is the encryption key used to encrypt the refresh token
 	EncryptionKey string `json:"encryption-key" yaml:"encryption-key" usage:"encryption key used to encryption the session state" env:"ENCRYPTION_KEY"`
 
-	// InvalidAuthRedirectsWith303 will make requests with invalid auth headers redirect using HTTP 303 instead of HTTP 307.  See github.com/keycloak/keycloak-gatekeeper/issues/292 for context.
-	InvalidAuthRedirectsWith303 bool `json:"invalid-auth-redirects-with-303" yaml:"invalid-auth-redirects-with-303" usage:"use HTTP 303 redirects instead of 307 for invalid auth tokens"`
 	// NoRedirects informs we should hand back a 401 not a redirect
 	NoRedirects bool `json:"no-redirects" yaml:"no-redirects" usage:"do not have back redirects when no authentication is present, 401 them"`
 	// SkipTokenVerification tells the service to skipp verifying the access token - for testing purposes
--- a/handlers.go
+++ b/handlers.go
@@ -101,7 +101,7 @@ func (r *oauthProxy) oauthAuthorizationHandler(w http.ResponseWriter, req *http.
 		return
 	}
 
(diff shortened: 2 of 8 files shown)


ackerleytng:
Thanks @abstractj !

Didn't know something similar was mentioned in the [RFC you quoted](https://tools.ietf.org/id/draft-ietf-oauth-security-topics-08.html#rfc.section.3.9)! 

Although strictly speaking, I think section 3.9 refers to the redirect from the authorization server back to the client (whose role louketo takes, in this case), so it's slightly different from the changes in the code in this PR, which is about redirects from the client to the authorization server in the earlier phase of the auth code flow prior to authentication with the authorization server.

I think the concerns still apply, though - we shouldn't be leaking information posted to the client to the authorization server either.

Also, I was confused by the header of this PR, I think you meant "Use HTTP 303 for redirects, instead of HTTP 307".

Thanks for your work on louketo! It's an awesomely useful proxy.

abstractj:
Hi @ackerleytng thanks for reviewing it. Indeed the title was incorrect and should be fixed now. Reading your last comments, what exactly is still being leaked now that we are using HTTP 303?

ackerleytng:
Thanks again @abstractj ! I think there's some misunderstanding there. I meant that the concerns expressed in the RFC are for redirection from authorisation server back to client, but the same concerns apply for us, even though we're redirecting from client to authorization server. 

Using a 303 should address those concerns! Thanks for making this change.

PR #630, opened 2020-06-02
Update OAuth2 dependency #518

### LINKED ISSUE #518: Update OAuth2 dependency

abstractj:
- package "crypto/rand" is being imported twice
- Remove unused reference
- Upgrade to golang/x/oauth2

Resolves #518


abstractj:
@jangaraj would you like to review?

ASzc:
Starting to have a look at this today

Issue #546, opened 2020-04-28
Add authorization services support to Gatekeeper


abstractj:
## What: 

  As a Gatekeeper developer, I want to enable a policy enforcer to my service or application so that I can enforce access to protected resources using fine-grained permissions.
  
  This issue is all about enabling policy-enforcer configuration, as we have in other adapters, to Gatekeeper.

## Reference:

- [KEYCLOAK-7502 - Add authorization services support to Gatekeeper](https://issues.redhat.com/browse/KEYCLOAK-7502)

p53:
i checked how it is done for application adapters and i think it is quite static, ideally there would be some context providers which would automatically generate body of request for permission ticket from parameters of incoming HTTP request e.g. keycloak_http_method_context_provider() would generate permission ticket request with this body:

```
'[
  {
    "resource_id": "{resource_id}",
    "resource_scopes": [
      "GET"
    ]
  }
]'
```
as request_id is passed in permission ticket it should be gathered from keycloak server e.g. by this request:

```
To query resources given an uri, send an HTTP GET request as follows:

http://${host}:${port}/auth/realms/${realm_name}/authz/protection/resource_set?uri=/api/alice
```

and if these providers would be stackable/additive e.g. something like:

authorized_context.add(keycloak_http_method_context_provider)
authorized_context.add(keycloak_http_headers_context_provider)

keycloak.create_permission_ticket(authorized_context)

or something like that

etc...

stianst:
There's one thing adding support for UMA 2, but not sure that fully covers Keycloak authorization services.

We also probably want the authz layer to be extensible I can see at least the following options in the future:

* Basic DSL built into Louketo
* WebHook - some clean/simple webhook API to plugin something external
* UMA 2
* Custom Go
* Keycloak Authorization Services - not sure if this is needed or we can do what we need to through UMA 2, or even supporting the Louketo defined WebHook API in Keycloak may be better

p53:
https://github.com/grepplabs/kafka-proxy, maybe this could be inspiration

pedroigor:
@stianst I would not consider UMA 2 for now. It solves a very specific set of use cases that are beyond API/HTTP security.

[...]

PR #673, opened 2020-07-23
#546 Change to support Keycloak Authz service

belint:
## Summary 
Louketo can do a request to Keycloak to check authorization policies for a resource.  Louketo will receive the decision from the application of policies then accept or not the user to access the asked resource.


sachinjohar2:
@belint Is this feature already available in keycloak-gatekeeper?

PR #677, opened 2020-08-07
Use ID token on refresh if access token is not a JWT

Yogu:
# Use ID token on refresh if access token is not a JWT

## Summary 

Currently, the refresh flow assumes that the access token is a JWT and if it isn't, the refresh fails. The initial token fetch routine already copes with this case by parsing the id token first, and only then tries to parse the access token. I copied the same logic into the token refresh routine.

## Type

[x] Bug fix
[] Feature request
[] Enhancement
[] Docs

## Why?

As far as I understand the OpenID Connect spec, you are not really supposed to parse the access token, and e.g. in the case of GitLab, it's not a JWT.

## Requirements

Having an OIDC provider that sends access tokens that are not JWTs

## How to try it?

Set up a GitLab application and use it, then see if the refresh flow works.

## Checklist:

- [x] My change requires a change to the documentation or CHANGELOG.
- [ ] I have updated the documentation/CHANGELOG accordingly.

### DIFF
--- a/oauth.go
+++ b/oauth.go
@@ -82,9 +82,21 @@ func getRefreshedToken(conf *oauth2.Config, t string) (jose.JWT, string, time.Ti
 		return jose.JWT{}, "", time.Time{}, time.Duration(0), err
 	}
 	refreshExpiresIn := time.Until(tkn.Expiry)
-	token, identity, err := parseToken(tkn.AccessToken)
+
+	rawIDToken, ok := tkn.Extra("id_token").(string)
+	if !ok {
+		return jose.JWT{}, "", time.Time{}, time.Duration(0), errors.New("unable to obtain id token")
+	}
+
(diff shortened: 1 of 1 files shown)


oauth.go
```diff
@@ -82,11 +82,23 @@ func getRefreshedToken(conf *oauth2.Config, t string) (jose.JWT, string, time.Ti
 		return jose.JWT{}, "", time.Time{}, time.Duration(0), err
 	}
 	refreshExpiresIn := time.Until(tkn.Expiry)
-	token, identity, err := parseToken(tkn.AccessToken)
+
+	rawIDToken, ok := tkn.Extra("id_token").(string)
+	if !ok {
+		return jose.JWT{}, "", time.Time{}, time.Duration(0), errors.New("unable to obtain id token")
```
Yogu, line 88: I guess we don't need to fail here if we still can parse the access token

